PR #155, opened 2022-02-09
Feature/apikeys

dikkeduif:


package.json
```diff
@@ -11,9 +11,9 @@
     "clean": "rm -rf ./dist tsconfig.tsbuildinfo",
     "compile": "tsc",
     "build": "yarn clean; yarn compile",
-    "start": "yarn migrate up; node -r module-alias/register ./dist/index.js",
+    "start": "yarn migrate up; node --inspect -r module-alias/register ./dist/index.js",
```
georgeroman, line 14: Are there any implications of running in debug mode in production?

src/api/routes.ts
```diff
@@ -11,6 +11,8 @@ import * as statsEndpoints from "@/api/endpoints/stats";
 import * as tokensEndpoints from "@/api/endpoints/tokens";
 import * as transfersEndpoints from "@/api/endpoints/transfers";
 import * as healthEndpoints from "@/api/endpoints/health";
+import * as apikeyEndpoints from "@/api/endpoints/apikeys"
+import { postApiKey } from '@/api/endpoints/apikeys';
```
georgeroman, line 15: Probably this line can get removed, since the one above covers it.

src/migrations/1644318693434_apikeys.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate';
+
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable("apikeys", {
```
georgeroman, line 7: For consistency with the rest of the tables, maybe let's rename this to `api_keys`.

src/migrations/1644318693434_apikeys.ts
```diff
@@ -0,0 +1,40 @@
...
+    key: {
+      type: "character(36)",
```
georgeroman, line 9: Let's change all `character` types to `text`. There is no performance difference in Postgres, and most people seem to recommend using `text`.

PR #178, opened 2022-02-27
Feature/volumes per day

dikkeduif:


dikkeduif:
I haven't added any additional indexes or filters on the collections table, lets maybe talk about what we want to do there. I don't see the same indexes as on the other branch.

PR #200, opened 2022-03-10
fix: optimize get user collections api

ipeleg:


georgeroman:
Nice, looks good to me, maybe we should put the changes into a new version (in case anyone depends on the old sorting order)?

PR #204, opened 2022-03-11
feat: created v3 sales api

ipeleg:


georgeroman:
I think it would be good to abstract out any pagination logic from the client. So instead of explicitly returning the fields that the pagination is done on in the API, maybe better to return them as a continuation, see below an example: https://github.com/reservoirprotocol/indexer-v3/blob/8d5858ff4089cef9bab8a5874a1f8fd9cf035ee1/src/api/endpoints/orders/get-orders-all/v1.ts#L127-L137

That API is paginated by (created_at, order_id) and so we return a continuation of `${createdAt}_${orderId}`. Then on the backend, whenever we receive a continuation we split that up into the individual fields. We could do the same with the sales API pagination. The nice thing about having such a continuation is that we abstract out the pagination details from the client (and can potentially change the keyset without breaking anything).

ipeleg:
Make sense, added the continuation node for easier pagination

src/api/endpoints/transfers/get-sales/v3.ts
```diff
@@ -0,0 +1,184 @@
...
+            .pattern(/^0x[a-f0-9]{40}$/),
+          amount: Joi.string(),
+          txHash: Joi.string()
+            .lowercase()
+            .pattern(/^0x[a-f0-9]{64}$/),
+          timestamp: Joi.number(),
+          price: Joi.number().unsafe().allow(null),
+          block: Joi.number(),
+          log_index: Joi.number(),
+          batch_index: Joi.number(),
```
georgeroman, line 63: Can keep `block` since that might be relevant, but let's remove `log_index` and `batch_index` since these are internal fields that are of no interest to any external clients.

PR #211, opened 2022-03-16
Feature/token continuation

dikkeduif:


src/api/endpoints/tokens/get-tokens-details/v2.ts
```diff
@@ -0,0 +1,411 @@
...
+
+import { Request, RouteOptions } from "@hapi/hapi";
+import Joi from "joi";
+
+import { edb } from "@/common/db";
+import { logger } from "@/common/logger";
+import { formatEth, fromBuffer, toBuffer } from "@/common/utils";
+import { getOrderSourceMetadata } from "@/orderbook/orders/utils";
+
+const version = "v1";
```
georgeroman, line 11: We should change the version to `v2`.

src/api/endpoints/tokens/get-tokens-details/v2.ts
```diff
@@ -0,0 +1,411 @@
...
+                  OR (t.top_buy_value is null)
+                 `);
+                (query as any).topBuyValue =
+                  contArr[0] !== "null"
+                    ? contArr[0]
+                    : "1000000000000000000000000";
+                (query as any).tokenId = contArr[1];
+              } else {
+                conditions.push(
+                  `(t.top_buy_value is null AND t.token_id > $/tokenId/)`
```
georgeroman, line 258: Shouldn't this be `token_id < $/tokenId/` since the index is on `token_id DESC`?

src/api/endpoints/tokens/get-tokens-details/v2.ts
```diff
@@ -0,0 +1,411 @@
...
+            case "topBidValue":
+              if (contArr[0] !== "null") {
+                conditions.push(`
+                  ("t"."top_buy_value", "t"."token_id") < ($/topBuyValue:raw/, $/tokenId:raw/)
+                  OR (t.top_buy_value is null)
+                 `);
+                (query as any).topBuyValue =
+                  contArr[0] !== "null"
+                    ? contArr[0]
+                    : "1000000000000000000000000";
```
georgeroman, line 254: What's this needed for? This case should never happen, right?

src/api/endpoints/tokens/get-tokens-details/v2.ts
```diff
@@ -0,0 +1,411 @@
...
+              if (contArr[0] !== "null") {
+                conditions.push(`(
+                  (t.floor_sell_value, "t"."token_id") > ($/floorSellValue/, $/tokenId/)
+                  OR (t.floor_sell_value is null)
+                )
+                `);
+                (query as any).floorSellValue =
+                  contArr[0] !== "null"
+                    ? contArr[0]
+                    : "1000000000000000000000000";
```
georgeroman, line 274: Same here as above.

src/api/endpoints/tokens/get-tokens/v2.ts
```diff
@@ -0,0 +1,295 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { Request, RouteOptions } from "@hapi/hapi";
+import Joi from "joi";
+
+import { edb } from "@/common/db";
+import { logger } from "@/common/logger";
+import { formatEth, fromBuffer, toBuffer } from "@/common/utils";
+
+const version = "v1";
```
georgeroman, line 10: Same as above, we should increase the version to know in which version requests/errors are happening.

georgeroman:
@dikkeduif Lgtm, let's merge this

georgeroman:
One last minor change, we should switch to using the cached attributes in the tokens table when filtering by attribute - just added that. Here's an example from the tokens/details API: https://github.com/reservoirprotocol/indexer-v3/blob/db520bcf0d21727c9579e5a13621f217249c37e4/src/api/endpoints/tokens/get-tokens-details/v1.ts#L207-L214.

PR #216, opened 2022-03-18
V5 peter

ptrwtts:
Hey @georgeroman here is a complete reorg of the docs, to make it more cohesive. I also made the following changes:

- Changed the routes for attribute APIs (and deprecated old routes)
- Created a "bulk" sales API that supports more results, but less data

Can you make sure everything looks good, and also make a few related changes:

1) Support querying the `collections` API by slug. This is important for fast lookups of slug into collection id
2) We need to support sample images for lists of collections, as multiple apps use this on V4. Can this be added to `collections` API with minimal cost? If so, let's do that. If not, we should change the `collectionOrSlug` API to be `collections/details` and support multiple results, similar to what we do for tokens
3) Get rid of the 0 and 99999999 defaults in floor asks API, so they don't appear in docs
4) Add an index on sales to allow getting all sales by date desc without a filter, which I added support for in the bulk sales API


src/api/routes.ts
```diff
@@ -276,6 +288,12 @@ export const setupRoutes = (server: Server) => {
     options: transfersEndpoints.getSalesV3Options,
   });
 
+  server.route({
+    method: "GET",
+    path: "/sales/bulk/v3",
```
georgeroman, line 293: Should this be v1?

PR #526, opened 2022-04-30
build: fix issues that prevent building locally

jrfrantz:
I tried to build and run the indexer locally and hit two issues. I found what they were and fixed them in this PR. Now, `docker-compose up` and `yarn build && yarn start` works on my machine.

1. Got an issue because hstore didn't exist, so I added it to the only sql file that i saw that used it.
2. Then I got a duplicate relation error for an index. I saw it was listed in two places so I got rid of one of them.

Don't feel strongly about landing this; putting up in case helpful.

### DIFF
--- a/src/migrations/1645976283000_daily-volumes.sql
+++ b/src/migrations/1645976283000_daily-volumes.sql
@@ -35,9 +35,6 @@ ALTER TABLE "collections"
 ALTER TABLE "collections"
   ADD "all_time_rank" INT;
 
-CREATE INDEX "fill_events_2_timestamp_index"
-  ON "fill_events_2" ("timestamp");
-
 CREATE INDEX "collections_day1_volume_index"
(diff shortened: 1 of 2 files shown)


src/migrations/1645976283000_daily-volumes.sql
```diff
@@ -35,9 +35,6 @@ ALTER TABLE "collections"
 ALTER TABLE "collections"
   ADD "all_time_rank" INT;
 
-CREATE INDEX "fill_events_2_timestamp_index"
```
jrfrantz, line 38: https://github.com/reservoirprotocol/indexer/blob/69ed6616df2de28e255cecf75e351b7d19e11545/src/migrations/1644930748093_fills.sql#L30

georgeroman:
Looks great, thanks for the contribution!

PR #674, opened 2022-05-17
added support for nft balances acquired_at

nofir:
- added `acquired_at` column on `nft_balances` table
- added logic to populate value when processing nft transfer events
- added support in get user tokens v2 api to expose acquired_at and support sorting
- added script to backfill acquired_at

### DIFF
--- a/src/api/endpoints/tokens/get-user-tokens/v2.ts
+++ b/src/api/endpoints/tokens/get-user-tokens/v2.ts
@@ -42,6 +42,8 @@ export const getUserTokensV2Options: RouteOptions = {
         .description(
           "Filter to a particular contract, e.g. `0x8d04a8c79ceb0889bdd12acdf3fa9d207ed3ff63`"
         ),
+      sortBy: Joi.string().valid("acquiredAt"),
+      sortDirection: Joi.string().lowercase().valid("asc", "desc").default("desc"),
       offset: Joi.number().integer().min(0).max(10000).default(0),
       limit: Joi.number().integer().min(1).max(20).default(20),
     }),
@@ -70,6 +72,7 @@ export const getUserTokensV2Options: RouteOptions = {
             tokenCount: Joi.string(),
             onSaleCount: Joi.string(),
             floorAskPrice: Joi.number().unsafe().allow(null),
+            acquiredAt: Joi.string().allow(null),
           }),
         })
       ),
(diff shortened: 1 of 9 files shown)


src/jobs/nft-balances-updates/backfill-acquired-at-queue.ts
```diff
@@ -0,0 +1,105 @@
...
+      if (_.size(values) > 0) {
+        const columns = new pgp.helpers.ColumnSet(
+          ["contract", "token_id", "owner", "acquired_at"],
+          {
+            table: "nft_balances",
+          }
+        );
+
+        try {
+          const updateQuery = `UPDATE nft_balances AS nb
```
georgeroman, line 56: Would it be faster (and maybe safer in the sense that it's atomically done) to use a single query (something like `WITH x AS (...) UPDATE ... FROM RETURNING ...`)?

src/migrations/1652750891708_nft_balances_acquired_date.sql
```diff
@@ -0,0 +1,15 @@
+-- Up Migration
+
+ALTER TABLE nft_balances
+ADD COLUMN acquired_at TIMESTAMPTZ;
+
+CREATE INDEX "nft_balances_owner_acquired_at_index"
```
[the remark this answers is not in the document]
nofir: That makes sense. Im going to do the following
1. Remove the CREATE / DROP INDEX statements from the migration.
2. Deploy the PR.
3. Run the backfill script. 
4. Run the CREATE/DROP INDEX CONCURRENTLY command as an individual command.
5. Add the CREATE / DROP INDEX statements to the migration file.

The sort by acquired_at in the get user tokens v2 api is no utilized yet so it makes sense to me to do #3 before #4.

Let me know if that makes sense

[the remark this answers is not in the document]
georgeroman: Yes, that makes sense, the only concern is the backfill script which depends on the `acquired_at IS NULL` condition. Without the index it might be slow to pinpoint the rows that need to be updated.

src/jobs/nft-balances-updates/backfill-acquired-at-queue.ts
nofir: thanks for the suggestion, made the changes.

src/api/endpoints/tokens/get-user-tokens/v2.ts
```diff
@@ -100,9 +103,20 @@ export const getUserTokensV2Options: RouteOptions = {
       collectionFilter = `AND t.collection_id = $/collection/`;
     }
 
+    let sortByFilter = "";
+    switch (query.sortBy) {
+      case "acquiredAt": {
+        sortByFilter = `
+            ORDER BY
+              b.acquired_at ${query.sortDirection} NULLS LAST
```
georgeroman, line 111: Since the index is on `DESC NULLS LAST` I'm not sure if an ascending sort would be able to use it. Maybe we can remove the `NULLS LAST` from the index and query since we ensure at a higher-level in the code that no entries with `amount > 0` will have a null `acquired_at`?

src/migrations/1643716513646_balances.sql
```diff
@@ -21,6 +22,10 @@ CREATE INDEX "nft_balances_contract_owner_index"
   INCLUDE ("amount")
   WHERE ("amount" > 0);
 
+CREATE INDEX "nft_balances_owner_acquired_at_index"
+  ON "nft_balances" ("owner", "acquired_at" DESC NULLS LAST)
```
georgeroman, line 26: From the index as well (without `NULLS LAST`), and should be good to go.

georgeroman:
@nofir Could you please recreate this PR? Sorry, I messed up branching a bit: merged dev into a branch I wanted to merge to main, then realized it includes some unrelated changes that were only on dev and then reverted the merge. Since the merge commit was reverted the changes are not on main but this PR still got closed.

PR #932, opened 2022-06-16
Feat/add order source fill events

nofir:
Run before deployment:

```
ALTER TABLE fill_events_2            
ADD COLUMN order_source_id_int INT;
```

```
CREATE INDEX CONCURRENTLY "fill_events_2_created_at_tx_hash_index"
  ON "fill_events_2" ("created_at", "tx_hash", "log_index", "batch_index");
```


src/sync/events/index.ts
```diff
@@ -1627,6 +1656,13 @@ export const syncEvents = async (
         }
       }
 
+      // Add order source value for each fill.
```
georgeroman, line 1659: Let's maybe only run this when not backfilling, i.e. `if (!backfill) { ... }`. Since backfilling is a very slow operation, we probably don't want to make it even slower.

nofir: makes sense, i'll make the change.

PR #1002, opened 2022-06-23
feat(collections): filter sample_images when null

jrfrantz:
A while ago I noticed some blank images show up in in preview cards for a communities marketplace.

The underlying issue was that some sample_images were null -- eg `[image1, null, null, image4]`. 

the issue that caused the nulls was that reservoir was getting rate-limited.

feels like the sample_images can be anything from the collection (eg doesnt need to be any particular 4 images), so this diff just filters out nulls.

note that i only did this for endpoints that were not tagged as x-deprecated.

orig discussion identifying the nulls at
https://discord.com/channels/872790973309153280/926358989770473493/987074452594499644

### DIFF
--- a/src/api/endpoints/collections/get-collection/v2.ts
+++ b/src/api/endpoints/collections/get-collection/v2.ts
@@ -166,6 +166,7 @@ export const getCollectionV2Options: RouteOptions = {
           ARRAY(
             SELECT "t"."image" FROM "tokens" "t"
             WHERE "t"."collection_id" = "c"."id"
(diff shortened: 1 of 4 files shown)


jrfrantz:
don't feel strongly that this lands - just putting it up to preempt the problem for the next person in case it's helpful.

as far as testing, i didn't have a good way to repro -- i only tested that `yarn build` still works on my local machine + that i can run an instance w a local docker image

ipeleg:
Look good mate!

Thank you for contributing! 🙌

PR #1212, opened 2022-07-18
feat: support multiple values selected per attribute

tv3636:


src/api/endpoints/tokens/get-tokens/v4.ts
```diff
@@ -170,22 +170,23 @@ export const getTokensV4Options: RouteOptions = {
...
+            ? `'${attributes[i].value.join("', '")}'`
+            : attributes[i].value;
+
           baseQuery += `
             JOIN "token_attributes" "ta${i}"
               ON "t"."contract" = "ta${i}"."contract"
               AND "t"."token_id" = "ta${i}"."token_id"
               AND "ta${i}"."key" = $/key${i}/
-              AND "ta${i}"."value" = $/value${i}/
+              AND "ta${i}"."value" ${multipleSelection ? `IN ($/value${i}:raw/)` : `= $/value${i}/`}
```
georgeroman, line 189: Let's not use `:raw` since that could result in SQL injection (eg. someone creating a collection with attribute values which interpolated are valid SQL strings). We do have some examples in the code where we can pass a list of values and `pg-promise` will handle the conversion to an SQL list.

PR #1288, opened 2022-07-27
New Exchange: Rarible

diegoramosxyz:


src/sync/events/data/rarible.ts
```diff
@@ -0,0 +1,21 @@
+import { Interface } from "@ethersproject/abi";
+import { EventData } from "@/events-sync/data";
+
+export const match: EventData = {
+  kind: "rarible-match",
+  addresses: { ["0x9757F2d2b135150BBeb65308D4a91804107cd8D6".toLowerCase()]: true },
```
georgeroman, line 6: The hardcoded address was just for testing. Let's take the example of all the other events and use the address that's found within the SDK (eg. `Sdk.Rarible.Addresses.Exchange[config.chainId]`). You should upgrade the SDK to version `0.0.133` (eg. `yarn add @reservoir0x/sdk@0.0.133`) before doing that.

src/sync/events/index.ts
```diff
@@ -1656,6 +1656,72 @@ export const syncEvents = async (
...
+                // Skip if the payment token is not supported.
+                break;
+              }
+
+              // Exclude orders with exotic asset types
+              if (!assetTypes.includes(leftAsset[0]) || !assetTypes.includes(rightAsset[0])) {
+                break;
+              }
+
+              const side = ["0x73ad2146", "0x973bb640"].includes(leftAsset[0]) ? "sell" : "buy";
```
georgeroman, line 1692: Let's also add a comment: `// Assume the left order is the maker's order`

georgeroman (changes requested):
Also, let's create a new PR which merges to `v5-dev` so we can backfill live on rinkeby and see whether we missed any edge cases before going for production deployment.

src/sync/events/index.ts
```diff
@@ -1656,6 +1656,72 @@ export const syncEvents = async (
...
+              const newRightFill = args["newRightFill"].toString();
+              const leftAsset = args["leftAsset"];
+              const rightAsset = args["rightAsset"];
+
+              // Keccak-256 hash
+              // ERC20 = 8ae85d84
+              // ETH = aaaebeba
+              // ERC721 = 73ad2146
+              // ERC1155 = 973bb640
+              const assetTypes = ["0x73ad2146", "0x973bb640", "0x8ae85d84", "0xaaaebeba"];
```
georgeroman, line 1675: Let's refactor these random bytes into variables (eg. `const ETH = "0xaaaebeba"`) for improved readability.

PR #1329, opened 2022-07-29
Element

diegoramosxyz:


src/sync/events/index.ts
```diff
@@ -1677,6 +1677,74 @@ export const syncEvents = async (
...
+
+            case "element-erc721-sell-order-filled": {
+              const { args } = eventData.abi.parseLog(log);
+              const maker = args["maker"].toLowerCase();
+              const taker = args["taker"].toLowerCase();
+              const erc20Token = args["erc20Token"].toLowerCase();
+              const erc20TokenAmount = args["erc20TokenAmount"].toString();
+              const erc721Token = args["erc721Token"].toLowerCase();
+              const erc721TokenId = args["erc721TokenId"].toString();
+              const orderHash = args["orderHash"];
```
georgeroman, line 1689: Let's call `toLowerCase()` on the order hash to make sure it's a lowercased hash (same below).

src/migrations/1659029948930_element-order-kind.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+
+ALTER TYPE "order_kind_t" ADD VALUE 'element';
```
georgeroman, line 3: Let's use the same convention as with zeroex-v4 orders, two types:
- element-erc721
- element-erc1155

src/sync/events/data/element.ts
```diff
@@ -0,0 +1,40 @@
...
+      address maker,
+      address taker,
+      address erc20Token,
+      uint256 erc20TokenAmount,
+      address erc721Token,
+      uint256 erc721TokenId,
+      bytes32 orderHash
+    )`,
+  ]),
+};
```
georgeroman, line 40: We should also handle ERC1155 events/orders, the ERC1155OrdersFeature in https://docs.element.market/developer/deployed-contract.

georgeroman (changes requested):

src/sync/events/index.ts
```diff
@@ -1677,6 +1677,140 @@ export const syncEvents = async (
...
+              const { args } = eventData.abi.parseLog(log);
+              const maker = args["maker"].toLowerCase();
+              const taker = args["taker"].toLowerCase();
+              const erc20Token = args["erc20Token"].toLowerCase();
+              const erc20TokenAmount = args["erc20TokenAmount"].toString();
+              const erc721Token = args["erc721Token"].toLowerCase();
+              const erc721TokenId = args["erc721TokenId"].toString();
+              const orderHash = args["orderHash"].toLowerCase();
+
+              const ETH = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee";
```
georgeroman, line 1691: Let's use `Sdk.ZeroExV4.Addresses.Eth[config.chainId]` here and in the other places where you define the `ETH` constant.

src/sync/events/index.ts
```diff
@@ -1677,6 +1677,140 @@ export const syncEvents = async (
...
+                break;
+              }
+
+              fillEventsPartial.push({
+                orderKind: "element-erc1155",
+                orderId: orderHash,
+                orderSide: "buy",
+                maker,
+                taker,
+                price: erc20FillAmount,
```
georgeroman, line 1805: For ERC1155 orders, I just pushed a fix to set the price of the fill event as the price of an individual item instead of the price of the whole batch, let's do the same for element: https://github.com/reservoirprotocol/indexer/blob/42e056e5bd0aa7aaecc29be9dd6fd0274c3a9348/src/sync/events/index.ts#L1362-L1363

georgeroman (changes requested):

PR #1385, opened 2022-08-03
Chore/polygon

nofir:


vdovaT:
Great!

src/sync/events/index.ts
```diff
@@ -59,6 +59,26 @@ export const syncEvents = async (
...
+      _.range(fromBlock, toBlock + 1).map((block) =>
+        limit(() => baseProvider.getBlockWithTransactions(block))
+      )
+    );
+
+    // Only allow attribution logic if we prefetched all blocks for efficiency.
+    handleAttribution = true;
+  }
+
   // Before proceeding, fetch all individual blocks within the current range
```
georgeroman, line 82: We should remove this since it's duplicated above

nofir: Fixed

PR #1395, opened 2022-08-04
Zora

diegoramosxyz:


src/sync/events/data/zora.ts
```diff
@@ -0,0 +1,24 @@
+import { Interface } from "@ethersproject/abi";
+import { Zora } from "@reservoir0x/sdk";
+import { config } from "@/config/index";
+import { EventData } from "@/events-sync/data";
+
+export const askFilled: EventData = {
+  kind: "zora-ask-filled",
+  addresses: { [Zora.Addresses.Exchange[config.chainId].toLowerCase()]: true },
```
georgeroman, line 8: Let's change this to `Zora.Addresses.Exchange[config.chainId]?.toLowerCase()` otherwise instances for which there is no defined exchange contract will error.

src/sync/events/index.ts
```diff
@@ -2077,6 +2077,38 @@ export const syncEvents = async (
...
+              const tokenContract = args["tokenContract"].toLowerCase();
+              const tokenId = args["tokenId"].toString();
+              const buyer = args["buyer"].toLowerCase();
+              const ask = args["ask"];
+
+              const seller = ask["seller"].toLowerCase();
+              const askCurrency = ask["askCurrency"].toLowerCase();
+              const askPrice = ask["askPrice"].toString();
+
+              if (![Sdk.Common.Addresses.Weth[config.chainId], AddressZero].includes(askCurrency)) {
```
georgeroman, line 2092: Should be consistent here and use `Sdk.Common.Addresses.Eth[config.chainId]` instead of `AddressZero`.

src/sync/events/index.ts
```diff
@@ -2077,6 +2077,38 @@ export const syncEvents = async (
...
+                // Skip if the payment token is not supported.
+                break;
+              }
+
+              fillEventsPartial.push({
+                orderKind: "zora",
+                currency: askCurrency,
+                orderSide: "buy",
```
georgeroman, line 2100: This is the `Asks` module, which implies listings and so the side should be 'sell'.

src/sync/events/index.ts
```diff
@@ -2077,6 +2077,38 @@ export const syncEvents = async (
...
+              const seller = ask["seller"].toLowerCase();
+              const askCurrency = ask["askCurrency"].toLowerCase();
+              const askPrice = ask["askPrice"].toString();
+
+              if (![Sdk.Common.Addresses.Weth[config.chainId], AddressZero].includes(askCurrency)) {
+                // Skip if the payment token is not supported.
+                break;
+              }
+
+              fillEventsPartial.push({
```
georgeroman, line 2097: No need to use `fillEventsPartial` here, can use `fillEvents`.

src/migrations/1659625683558_zora-order-kind.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+
+ALTER TYPE "order_kind_t" ADD VALUE 'zora';
```
georgeroman, line 3: Let's use `zora-v3` for the name.

georgeroman (changes requested):

georgeroman:
Also, might need to rebase your branch on `v5` and incorporate some of the latest changes there like handling attribution and setting the correct prices in ETH and USD.

georgeroman:
Also, let's handle `zora-v3` as a default source: https://github.com/reservoirprotocol/indexer/blob/5eb39ce3ca4b3cd076f1935349c28d820513a838/src/sync/events/utils.ts#L103-L124

georgeroman:
We recently added support for handling sales in ERC20 tokens (anything other than ETH/WETH), so let's also make sure we properly handle those here. You can see here an example on how to deal with it: https://github.com/reservoirprotocol/indexer/blob/562724971a658d8faaca53e0382de2dd2d08e9cd/src/sync/events/index.ts#L610-L636.

diegoramosxyz:
@georgeroman Why do you have a check for WETH and ETH here:

https://github.com/reservoirprotocol/indexer/blob/562724971a658d8faaca53e0382de2dd2d08e9cd/src/sync/events/index.ts#L560-L568

Isn't the `getPrices` function already making this check? Wouldn't this check nullify checks for other ERC20s?

georgeroman:
Looks good, let's resolve the conflicts and should be good to merge.

PR #1398, opened 2022-08-04
Cryptopunks

diegoramosxyz:


georgeroman:
The punks market might require some hacks due to this issue: https://github.com/larvalabs/cryptopunks/issues/19. We basically need to index their `Transfer` events as well in order to correctly set the bugged field on the `PunkBought` event. We should also handle the punks market as a default source and update the pricing (check the comments on the zora PR).

ptrwtts:
We should probably set the source to `cryptopunks.app`, which is the new domain for the marketplace since Yuga bought it.

src/sync/events/index.ts
```diff
@@ -2077,6 +2077,47 @@ export const syncEvents = async (
...
+              break;
+            }
+
+            case "cryptopunks-transfer": {
+              const { args } = eventData.abi.parseLog(log);
+              const from = args["from"].toLowerCase();
+              const to = args["to"].toLowerCase();
+              const value = args["value"].toString();
+
+              nftTransferEvents.push({
```
georgeroman, line 2110: For the moment, let's not save these transfer events in the database.

src/sync/events/index.ts
```diff
@@ -2077,6 +2077,47 @@ export const syncEvents = async (
...
+                contract: Sdk.CryptoPunks.Addresses.Exchange[config.chainId]?.toLowerCase(),
+                tokenId: punkIndex,
+                amount: "1",
+                baseEventParams,
+              });
+
+              break;
+            }
+
+            case "cryptopunks-transfer": {
```
georgeroman, line 2104: Simply indexing the transfer events is not going to fix our problem, we need to use the transfer event for properly setting the `toAddress`. Also, since the `value` filled is bugged as well, we need to find a way to properly set that as well (a quick workaround would be to parse the `minPrice` parameter from the calldata and use that as the price. We do similar processing for handling wyvern sales, so can use that code as an example: https://github.com/reservoirprotocol/indexer/blob/562724971a658d8faaca53e0382de2dd2d08e9cd/src/sync/events/index.ts#L1077-L1106

georgeroman (changes requested):

src/sync/events/index.ts
```diff
@@ -2077,6 +2086,77 @@ export const syncEvents = async (
...
+              const { args } = eventData.abi.parseLog(log);
+              const from = args["from"].toLowerCase();
+              const to = args["to"].toLowerCase();
+              const value = args["value"].toString();
+
+              cryptopunksTransferEvents.push({
```
georgeroman, line 2149: We only care about `to` and `txHash`, let's only store these.

src/sync/events/index.ts
```diff
@@ -2077,6 +2086,77 @@ export const syncEvents = async (
...
+
+              fillEventsPartial.push({
+                orderKind: "cryptopunks",
+                orderSide: "buy",
+                taker: fromAddress,
+                maker: toAddress,
+                price: prices.nativePrice,
+                usdPrice: prices.usdPrice,
+                currency: AddressZero,
+                contract: Sdk.CryptoPunks.Addresses.Exchange[config.chainId]?.toLowerCase(),
```
georgeroman, line 2134: Let's use `baseEventParams.address`.

src/sync/events/index.ts
```diff
@@ -2077,6 +2086,77 @@ export const syncEvents = async (
...
+              if (!prices.nativePrice) break;
+
+              fillEventsPartial.push({
+                orderKind: "cryptopunks",
+                orderSide: "buy",
+                taker: fromAddress,
+                maker: toAddress,
+                price: prices.nativePrice,
+                usdPrice: prices.usdPrice,
+                currency: AddressZero,
```
georgeroman, line 2133: Let's use the `Sdk.Common.Addresses.Eth[config.chainId]`.

src/sync/events/index.ts
```diff
@@ -2077,6 +2086,77 @@ export const syncEvents = async (
...
+                value,
+                baseEventParams.timestamp
+              );
+
+              // We must always have the native price
+              if (!prices.nativePrice) break;
+
+              fillEventsPartial.push({
+                orderKind: "cryptopunks",
+                orderSide: "buy",
```
georgeroman, line 2128: The order side could also be `sell`, we can actually use a little hack and depending on the `toAddress` (whether it's `0x00.0..000` or not) set the side to `buy` or `sell` (eg. `buy` if it is the zero address, `sell` otherwise).

src/sync/events/index.ts
```diff
@@ -2077,6 +2086,77 @@ export const syncEvents = async (
...
             }
+
+            case "cryptopunks-punk-bought": {
+              const { args } = eventData.abi.parseLog(log);
+              const punkIndex = args["punkIndex"].toString();
+              let value = args["value"].toString();
+              const fromAddress = args["fromAddress"].toLowerCase();
+              let toAddress = args["toAddress"].toLowerCase();
+
+              if (
```
georgeroman, line 2097: Let's maybe add some comments explaining why we do these hacks the way we do them.

georgeroman (changes requested):

src/sync/events/index.ts
```diff
@@ -2077,6 +2082,88 @@ export const syncEvents = async (
...
+              const prices = await getPrices(
+                Sdk.Common.Addresses.Eth[config.chainId],
+                value,
+                baseEventParams.timestamp
+              );
+
+              // We must always have the native price
+              if (!prices.nativePrice) break;
+
+              const orderSide = toAddress === AddressZero ? "buy" : "sell";
```
georgeroman, line 2134: This is not going to work since we already fix the zero address issue above. We should have this line before we make the fix.

src/sync/events/index.ts
```diff
@@ -2077,6 +2082,88 @@ export const syncEvents = async (
...
+              }
+
+              const prices = await getPrices(
+                Sdk.Common.Addresses.Eth[config.chainId],
+                value,
+                baseEventParams.timestamp
+              );
+
+              // We must always have the native price
+              if (!prices.nativePrice) break;
```
georgeroman, line 2132: Let's add `{ ... }` for consistency with the rest of the codebase.

PR #1455, opened 2022-08-15
feat: remove burned tokens from stats

jrfrantz:
This PR attempts to remove burned tokens (eg owned by `address(0)`) from the stats endpoint. I don't know how to test this locally so let me know if it looks off and happy to change whatever.

Discussion in discord was to remove burned tokens from stats but leave in other tokens endpoints (where they can be more easily removed on client). Link: https://discord.com/channels/872790973309153280/1007323955574603847/1007631754103898132

### DIFF
--- a/src/api/endpoints/stats/get-stats/v1.ts
+++ b/src/api/endpoints/stats/get-stats/v1.ts
@@ -9,6 +9,8 @@ import { logger } from "@/common/logger";
 import { formatEth, fromBuffer, toBuffer } from "@/common/utils";
 import { Tokens } from "@/models/tokens";
 
+import { AddressZero } from "@ethersproject/constants";
+
 const version = "v1";
 
 export const getStatsV1Options: RouteOptions = {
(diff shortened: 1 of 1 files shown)


src/api/endpoints/stats/get-stats/v1.ts
```diff
@@ -268,6 +274,11 @@ export const getStatsV1Options: RouteOptions = {
         `;
       } else if (query.collection) {
         baseQuery = `
+          WITH "burned" AS (
+            SELECT * FROM "nft_balances"
+            WHERE "owner" = ${toBuffer(AddressZero)}
```
georgeroman, line 279: String interpolation is not going to work, we need to have a new query param (eg. `burnAddress`) and set that to `toBuffer(AddressZero)` in the query params object.

src/api/endpoints/stats/get-stats/v1.ts
```diff
@@ -291,7 +302,7 @@ export const getStatsV1Options: RouteOptions = {
             LIMIT 1
           )
           SELECT
-            "c"."token_count",
+            "c"."token_count" - (SELECT COUNT(*) FROM "burned") AS "token_count",
```
georgeroman, line 305: No need for the `WITH` clause, could just have something like `(SELECT COUNT(*) FROM nft_balances WHERE contract = ... AND owner = ... AND amount > 0)`.

georgeroman (changes requested):

nofir:
closing due to lack of activity

PR #1495, opened 2022-08-16
Remove redundant checks

diegoramosxyz:
All ERC20 tokens are now supported, so it's no longer necessary to check whether the token is WETH or not.


georgeroman:
I wonder where these unrelated changes are coming from. Make sure to first `git pull origin main` on `main`, then branch out a new branch from that and as long as this is done, there shouldn't be that many conflicts/changes.

georgeroman:
Closing this, already took care of removing these redundant checks in a different PR.

PR #1524, opened 2022-08-19
feat: add fill/order source filter to sales/v3

zqjimlove:


georgeroman:
@ipeleg Can you take a look at this? My main concern is missing indexes on the sales table.

PR #1675, opened 2022-08-31
Add Universe

ivostoynovski:


src/sync/events/data/universe.ts
```diff
@@ -0,0 +1,38 @@
+import { Interface } from "@ethersproject/abi";
+import { Universe } from "@reservoir0x/sdk";
+import { config } from "@/config/index";
+import { EventData } from "@/events-sync/data";
+
+export const match: EventData = {
+  kind: "universe-match",
+  addresses: { [Universe.Addresses.Exchange[config.chainId]?.toLowerCase()]: true },
+  topic: "0x268820db288a211986b26a8fda86b1e0046281b21206936bb0e61c67b5c79ef4",
+  numTopics: 1,
```
georgeroman, line 10: This seems to be wrong, it should be `1 + numberOfIndexedEventParams`.

src/sync/events/data/universe.ts
```diff
@@ -0,0 +1,38 @@
...
+    `event Match(
+      bytes32 indexed leftHash,
+      bytes32 indexed rightHash,
+      address indexed leftMaker,
+      address rightMaker,
+      uint newLeftFill,
+      uint newRightFill,
+      (bytes4 assetClass, bytes data) leftAsset,
+      (bytes4 assetClass, bytes data) rightAsset
+    );`,
```
georgeroman, line 21: Let's remove the trailing `;` and replace `uint` with `uint256` for consistency with the rest of the code.

src/sync/events/data/universe.ts
```diff
@@ -0,0 +1,38 @@
...
+  ]),
+};
+
+export const cancel: EventData = {
+  kind: "universe-cancel",
+  addresses: { [Universe.Addresses.Exchange[config.chainId]?.toLowerCase()]: true },
+  topic: "0xbbdc98cb2835f4f846e6a63700d0498b4674f0e8858fd50c6379314227afa04e",
+  numTopics: 1,
```
georgeroman, line 29: Same as above, doesn't match the number of indexed event parameters.

src/sync/events/index.ts
```diff
@@ -2363,6 +2363,132 @@ export const syncEvents = async (
...
+                usdPrice: prices.usdPrice,
+                contract,
+                tokenId,
+                amount,
+                baseEventParams,
+              });
+
+              break;
+            }
+            case "universe-cancel": {
```
georgeroman, line 2467: Let's add a newline in-between.

src/sync/events/index.ts
```diff
@@ -2363,6 +2363,132 @@ export const syncEvents = async (
 
               break;
             }
+
+            case "universe-match": {
```
georgeroman, line 2367: Is this exactly the same as Rarible? In that case we should merge the logic for handling Rarible and Universe, similarly to how we do it for OpenDao/ZeroExV4 and WyvernV2/WyvernV23: https://github.com/reservoirprotocol/indexer/blob/c5ca1082c246f47b7e02780ee71ff51b0358db8e/src/sync/events/index.ts#L1163-L1164
https://github.com/reservoirprotocol/indexer/blob/c5ca1082c246f47b7e02780ee71ff51b0358db8e/src/sync/events/index.ts#L1339-L1340

georgeroman (changes requested):

src/sync/events/index.ts
```diff
@@ -1717,6 +1717,7 @@ export const syncEvents = async (
               break;
             }
 
+            case "universe-match":
             case "rarible-match": {
```
georgeroman, line 1721: That's a good start, but we still want to set a different order kind (line 1787) depending on whether the current event is a `rarible-match` or a `universe-match`.

georgeroman:
Closing this, manually ported the changes that were deployed to dev.

PR #1903, opened 2022-09-20
feat: zora orderbook support

lljxx1:
### Overview
- move zora releated code to `src/exchanges/zora`
- add jest for test eg: `test/zora-events.test.ts`
- tests coverd: `order create > update order`  and `order created > order canceled`
- order-check include: `balance-check` and `approval-check`

### File Structe
- `/__fixtures__` test resource
- `/orderbook/check.ts` order-check
- `/orderbook/index.ts` handle orders queue
- `/tests/*` test code
- `/event.ts` all events
- `/handlers` event parsing
- `/index.ts` exports


tsconfig.json
```diff
@@ -23,9 +23,13 @@
       "@/utils/*": ["utils/*"],
       "@/models/*": ["models/*"],
       "@/pubsub/*": ["pubsub/*"],
+      "@/exchanges/*": ["exchanges/*"],
```
georgeroman, line 26: For consistency with the rest of the code let's move the logic in `src/exchanges` to `src/sync/events` (the events syncing logic) and `src/orderbook/order` (the order persistence and validation logic).

src/sync/events/index.ts
```diff
@@ -264,135 +264,91 @@ export const syncEvents = async (
...
-
-            const price = bn(amount).add(curatorFee).toString();
-
-            // Handle: attribution
-
-            const orderKind = "zora-v3";
-            const data = await syncEventsUtils.extractAttributionData(
-              baseEventParams.txHash,
-              orderKind
+          // Quixotic
```
georgeroman, line 267: Could you rebase on top of `v5`? This part of the code got removed in some recent commits.

src/sync/events/handlers/index.ts
```diff
@@ -1,5 +1,6 @@
 import { EnhancedEvent, OnChainData, processOnChainData } from "@/events-sync/handlers/utils";
 
+import * as zora from "@/exchanges/zora/handler";
```
georgeroman, line 3: Let's move this import at the end, for consistency (same goes for the `kind` type definition and the switch statement).

src/sync/events/data/index.ts
```diff
@@ -19,7 +19,7 @@ import * as wyvernV2 from "@/events-sync/data/wyvern-v2";
 import * as wyvernV23 from "@/events-sync/data/wyvern-v2.3";
 import * as x2y2 from "@/events-sync/data/x2y2";
 import * as zeroExV4 from "@/events-sync/data/zeroex-v4";
-import * as zora from "@/events-sync/data/zora";
+import * as zora from "@/exchanges/zora/events";
```
georgeroman, line 22: As mentioned below, we should keep this logic in `events-sync` for consistency.

src/jobs/orderbook/orders-queue.ts
```diff
@@ -54,6 +55,14 @@ if (config.doBackgroundWork) {
             break;
           }
 
+          case "zora-v3": {
+            const result = await exchanges.zora.orders.save([
+              info as exchanges.zora.orders.OrderInfo,
+            ]);
+            logger.info(QUEUE_NAME, `[zora-3] Order save result: ${JSON.stringify(result)}`);
```
georgeroman, line 62: Typo: `zora-3` instead of `zora-v3`

src/exchanges/zora/orderbook/index.ts
```diff
@@ -0,0 +1,251 @@
...
+        } else {
+          return results.push({
+            id,
+            status: "not-fillable",
+          });
+        }
+      }
+
+      if (orderExists) {
+        // If an older order already exists then we just update some fields on it
```
georgeroman, line 90: One very important property of our indexer is that the order of syncing the events shouldn't matter. That is, you can sync block ranges 5-10, then 1-2 and then 3-4 and the events should be properly handled without any dependence on syncing them in their original order. The way this part of the code works is that it assumes to we're never going to sync order creation events that occurred in the past. However, that is wrong because we do backfills on a regular basis (backfill = syncing old events we previously missed / ignored / mishandled). As such, we should handle Zora orders in the same way we handle Foundation / CryptoPunks orders (see below) - only saving / updating an order if it is newer than the already existing order (otherwise, backfills might overwrite orders with old and stale data).
https://github.com/reservoirprotocol/indexer/blob/00eea91b535904ae88c64f287c31006c68c5fbaa/src/orderbook/orders/foundation/index.ts#L81-L129

src/exchanges/zora/orderbook/index.ts
```diff
@@ -0,0 +1,251 @@
...
+
+export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
+  const results: SaveResult[] = [];
+  const orderValues: DbOrder[] = [];
+
+  const handleOrder = async ({ orderParams, metadata }: OrderInfo) => {
+    try {
+      const id = getOrderId(orderParams);
+
+      // Check: order doesn't already exist
```
georgeroman, line 61: We can probably remove this `orderExists` once you make the change requested below.

src/exchanges/zora/orderbook/index.ts
```diff
@@ -0,0 +1,251 @@
...
+
+        return results.push({
+          id,
+          status: "success",
+        });
+      }
+
+      // Check and save: associated token set
+      const schemaHash = metadata.schemaHash ?? generateSchemaHash(metadata.schema);
+      const contract = Sdk.Zora.Addresses.Exchange[config.chainId];
```
georgeroman, line 127: The contract should be the token's contract and not the exchange contract.

src/exchanges/zora/orderbook/index.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const handleOrder = async ({ orderParams, metadata }: OrderInfo) => {
+    try {
+      const id = getOrderId(orderParams);
+
+      // Check: order doesn't already exist
+      const orderExists = await idb.oneOrNone(`SELECT 1 FROM "orders" "o" WHERE "o"."id" = $/id/`, {
+        id,
+      });
+
```
georgeroman, line 65: For Zora we should ignore non-ETH listings, as below:
https://github.com/reservoirprotocol/indexer/blob/00eea91b535904ae88c64f287c31006c68c5fbaa/src/orderbook/orders/zeroex-v4/index.ts#L140-L149

src/exchanges/zora/orderbook/index.ts
```diff
@@ -0,0 +1,251 @@
...
+        currency_price: orderParams.askPrice.toString(),
+        currency_value: orderParams.askPrice.toString(),
+        needs_conversion: null,
+        quantity_remaining: "1",
+        valid_between: `tstzrange(${validFrom}, ${validTo}, '[]')`,
+        nonce: null,
+        source_id_int: source?.id,
+        is_reservoir: null,
+        contract: toBuffer(contract),
+        conduit: null,
```
georgeroman, line 170: Since Zora orders are not escrowed (eg. the NFT being sold is not locked in the exchange contract), orders might get validated / invalidated due to balance / approval changes. The role of the conduit is to properly handle approval changes. Whenever an approval is revoked on a particular contract, any orders which have that contract as a conduit will get their approval status changes to 'no-approval', and viceversa for approval granting. The only requirement is that we properly set this conduit field to the contract that is given the NFT approval for Zora (I guess that's the transfer manager).

src/exchanges/zora/handler.ts
```diff
@@ -0,0 +1,257 @@
...
+  };
+}
+
+export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData> => {
+  const cancelEventsOnChain: es.cancels.Event[] = [];
+  const fillEventsOnChain: es.fills.Event[] = [];
+  const nftTransferEvents: es.nftTransfers.Event[] = [];
+
+  const fillInfos: fillUpdates.FillInfo[] = [];
+  const mintInfos: tokenUpdatesMint.MintInfo[] = [];
```
georgeroman, line 38: Let's remove everything we don't need (that is, `nftTransferEvents` and `mintInfos`).

src/exchanges/zora/handler.ts
```diff
@@ -0,0 +1,257 @@
...
+          askCurrency,
+          askPrice,
+          baseEventParams.timestamp
+        );
+        if (!prices.nativePrice) {
+          // We must always have the native price
+          break;
+        }
+
+        fillEventsOnChain.push({
```
georgeroman, line 79: We can/should attach an order id to the fill event, as below:
https://github.com/reservoirprotocol/indexer/blob/00eea91b535904ae88c64f287c31006c68c5fbaa/src/sync/events/handlers/foundation.ts#L91

src/exchanges/zora/events.ts
```diff
@@ -0,0 +1,112 @@
...
+      ) ask
+    )`,
+]);
+
+export const askCreated: EventData = {
+  kind: "zora-ask-created",
+  addresses: { [Zora.Addresses.Exchange[config.chainId]?.toLowerCase()]: true },
+  topic: askCreatedABI.getEventTopic("AskCreated"),
+  numTopics: 3,
+  abi: askCreatedABI,
```
georgeroman, line 67: We can store the ABI directly in the `abi` field instead of having a new variable for it, for consistency (same below).

georgeroman (changes requested):
A few high-level comments:
- for consistency with the rest of the code let's move the logic in `src/exchanges` to `src/sync/events` (the events syncing logic) and `src/orderbook/order` (the order persistence and validation logic)
- let's remove `package-lock.json` since we already have `yarn.lock`

src/orderbook/orders/zora/index.ts
```diff
@@ -0,0 +1,273 @@
...
+        currency_price: orderParams.askPrice.toString(),
+        currency_value: orderParams.askPrice.toString(),
+        needs_conversion: null,
+        quantity_remaining: "1",
+        valid_between: `tstzrange(${validFrom}, ${validTo}, '[]')`,
+        nonce: null,
+        source_id_int: source?.id,
+        is_reservoir: null,
+        contract: toBuffer(contract),
+        conduit: toBuffer(Sdk.Zora.Addresses.Erc721TransferHelper[config.chainId]),
```
georgeroman, line 192: Although we only support listings, let's have the following logic in order to make sure we don't miss it when we'll add bids:
`orderParams.side === "sell" ? Sdk.Zora.Addresses.Erc721TransferHelper[config.chainId] : Sdk.Zora.Addresses.Erc20TransferHelper[config.chainId]`

georgeroman (changes requested):
Let's also update the SDK to `0.0.174`, just added the zora addresses on goerli: https://github.com/reservoirprotocol/core/commit/5ba12f363c61856fc218d4d299bc487f5baf2b7a. We'll then deploy to dev and test there.

georgeroman:
Also, let's switch the merge branch to `v5-dev` instead of `v5` since we'll first deploy to dev.

src/sync/events/handlers/zora.ts
```diff
@@ -0,0 +1,256 @@
...
+            logIndex: baseEventParams.logIndex,
+            batchIndex: baseEventParams.batchIndex,
+            blockHash: baseEventParams.blockHash,
+          },
+        });
+
+        break;
+      }
+
+      case "zora-ask-price-updated": {
```
georgeroman, line 159: Just realized an edge-case here. When updating the price of an ask, a user can update the currency as well (see code below), but if the updated currency is not ETH then we will just ignore it (since we don't support non-ETH listings) and thus store the old listing which is not relevant anymore. The way to handle this is that if a listing is non-ETH AND the order already exists in the database then we should cancel the old order (set it's fillability status as 'cancelled').

Here's the Zora code for updating a listing:
https://github.com/ourzora/v3/blob/32289b6d549b8f5aa8e08e7f21d4de32a7b4a78b/contracts/modules/Asks/V1.1/AsksV1_1.sol#L188-L202

And here's the indexer code where we should add some additional logic to handle this edge case:
https://github.com/reservoirprotocol/indexer/blob/84eb164b9cad8d03cbf2002818be9a85cc5ba3ac/src/orderbook/orders/zora/index.ts#L61-L67

src/orderbook/orders/zora/index.ts
```diff
@@ -0,0 +1,277 @@
...
+    );
+    await idb.none(pgp.helpers.insert(orderValues, columns) + " ON CONFLICT DO NOTHING");
+
+    await ordersUpdateById.addToQueue(
+      results
+        .filter((r) => r.status === "success" && !r.unfillable)
+        .map(
+          ({ id }) =>
+            ({
+              context: `new-order-${id}`,
```
georgeroman, line 266: Also, let's use this context `new-order-${id}-${txHash}`, otherwise some jobs might get skipped (the context basically acts as a unique identifier for the job, so that if we ever put a job with the same context back on the queue it will get skipped since it knows it already processed a job with this context - in reality only the last 10000 jobs are kept in memory and compared against, so this issue is only relevant for a new ask and then an ask update which occur very close to each other).

PR #1904, opened 2022-09-20
Universe Indexer Integration

vikinatora:
Integrate Universe Orderbook into the Indexing process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@hapi/inert": "^6.0.4",
     "@hapi/vision": "^6.1.0",
     "@poprank/rankings": "^1.1.20",
-    "@reservoir0x/sdk": "0.0.182",
+    "@reservoir0x/sdk": "0.0.183",
     "@types/date-fns": "^2.6.0",
     "@types/hapi__basic": "^5.1.2",
     "@types/hapi__hapi": "^20.0.9",
--- a/src/api/endpoints/execute/get-execute-bid/v4.ts
+++ b/src/api/endpoints/execute/get-execute-bid/v4.ts
@@ -30,6 +30,9 @@ import * as zeroExV4BuyAttribute from "@/orderbook/orders/zeroex-v4/build/buy/at
 import * as zeroExV4BuyToken from "@/orderbook/orders/zeroex-v4/build/buy/token";
 import * as zeroExV4BuyCollection from "@/orderbook/orders/zeroex-v4/build/buy/collection";
 
+// Universe
+import * as universeBuyToken from "@/orderbook/orders/universe/build/buy/token";
+
 const version = "v4";
 
 export const getExecuteBidV4Options: RouteOptions = {
@@ -76,18 +79,18 @@ export const getExecuteBidV4Options: RouteOptions = {
             "Bid on a particular attribute value. Example: `Teddy (#33)`"
           ),
           quantity: Joi.number().description(
-            "Quanity of tokens user is buying. Only compatible with ERC1155 tokens. Example: `5`"
+            "Quantity of tokens user is buying. Only compatible with ERC1155 tokens. Example: `5`"
           ),
           weiPrice: Joi.string()
             .pattern(regex.number)
             .description("Amount bidder is willing to offer in wei. Example: `1000000000000000000`")
             .required(),
           orderKind: Joi.string()
-            .valid("zeroex-v4", "seaport", "looks-rare", "x2y2")
+            .valid("zeroex-v4", "seaport", "looks-rare", "x2y2", "universe")
             .default("seaport")
             .description("Exchange protocol used to create order. Example: `seaport`"),
           orderbook: Joi.string()
-            .valid("reservoir", "opensea", "looks-rare", "x2y2")
+            .valid("reservoir", "opensea", "looks-rare", "x2y2", "universe")
             .default("reservoir")
             .description("Orderbook where order is placed. Example: `Reservoir`"),
           automatedRoyalties: Joi.boolean()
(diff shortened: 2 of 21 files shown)


src/api/endpoints/execute/get-execute-bid/v4.ts
```diff
@@ -598,6 +602,89 @@ export const getExecuteBidV4Options: RouteOptions = {
             // Go on with the next bid
             continue;
           }
+
+          case "universe": {
+            if (!["universe"].includes(params.orderbook)) {
```
georgeroman, line 607: Do we want to allow posting directly to `reservoir` as well?

src/api/endpoints/execute/get-execute-list/v4.ts
```diff
@@ -511,6 +516,87 @@ export const getExecuteListV4Options: RouteOptions = {
               orderIndex: i,
             });
           }
+
+          // Not sure why eslint isn't happy
+          // eslint-disable-next-line no-fallthrough
```
georgeroman, line 521: I think we're missing the `continue` on the above case.

src/api/endpoints/execute/get-execute-sell/v4.ts
```diff
@@ -181,6 +180,7 @@ export const getExecuteSellV4Options: RouteOptions = {
           quantity: 1,
           source: sourceId ? sources.get(sourceId)?.domain ?? null : null,
           // TODO: Add support for multiple currencies
+          // TODO: Ask George, how we can support APE coin?
```
georgeroman, line 183: On mainnet for listings APE coin will be supported by default (as long as the SDK code supports listing in ERC20) - no need to do any special handling for it.

src/api/endpoints/execute/get-execute-bid/v4.ts
```diff
@@ -213,6 +216,7 @@ export const getExecuteBidV4Options: RouteOptions = {
           params.feeRecipient.push(feeRecipient);
         }
 
+        //TODO: Add support for more ERC20 tokens
```
georgeroman, line 219: We do have in plan bids in non-WETH tokens (probably USDC at first) - unfortunately due to how the indexer works (eg. it needs to have the full balance of every owner for properly validating the orders) we need an extra backfill for every ERC20 token we support on the bid side.

src/api/endpoints/execute/get-execute-buy/v4.ts
```diff
@@ -520,11 +520,12 @@ export const getExecuteBuyV4Options: RouteOptions = {
           throw Boom.badData("Balance too low to proceed with transaction");
         }
 
-        if (!listingDetails.every((d) => d.kind === "seaport")) {
-          throw new Error("Only Seaport ERC20 listings are supported");
+        if (!listingDetails.every((d) => d.kind === "seaport" || d.kind === "universe")) {
```
georgeroman, line 523: I think we either want all `seaport` or all `universe`. If everything is `seaport` then we keep the code below which gets the conduit, otherwise we need the conduit to be the universe exchange address (or whatever contract we need the approval on).

src/api/endpoints/execute/get-execute-list/v4.ts
```diff
@@ -511,6 +516,87 @@ export const getExecuteListV4Options: RouteOptions = {
...
+          case "universe": {
+            if (!["universe"].includes(params.orderbook)) {
+              throw Boom.badRequest("Only `reservoir` is supported as orderbook");
```
georgeroman, line 524: Given the if condition, should this be `universe`? Also, see the comment above on allowing `reservoir` as well.

src/orderbook/orders/universe/check.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  if (options?.checkFilledOrCancelled) {
+    // Check: order is not cancelled
+    const cancelled = await commonHelpers.isOrderCancelled(id);
+    if (cancelled) {
+      throw new Error("cancelled");
+    }
+
+    // Check: order is not filled
+    //TODO: Not sure if this will work. Ask George@Reservoir?
```
georgeroman, line 49: This should work, but probably better to rename `filled` to `amount` or something that better denotes that it's the total quantity the order was intended for.

src/sync/events/handlers/universe.ts
```diff
@@ -0,0 +1,187 @@
...
+
+        // Assume the left order is the maker's order
+        const side = [ERC721, ERC1155].includes(leftAsset.assetClass) ? "sell" : "buy";
+
+        const currencyAsset = side === "sell" ? rightAsset : leftAsset;
+        const nftAsset = side === "sell" ? leftAsset : rightAsset;
+
+        // Handle: attribution
+
+        const orderKind = eventData.kind.startsWith("universe") ? "universe" : "";
```
georgeroman, line 98: No need for the `? :` here, we can just set `orderKind = "universe"` directly.

src/sync/events/index.ts
```diff
@@ -367,6 +366,10 @@ export const syncEvents = async (
             kind === "erc20-transfer"
         ),
       },
+      {
+        kind: "universe",
```
georgeroman, line 370: Since we're already handling `universe` in its own handler file, let's remove it from above (and only keep `rarible-match`).

georgeroman (changes requested):

src/api/endpoints/execute/get-execute-bid/v4.ts
vikinatora: I will talk with the PO about this.

src/api/endpoints/execute/get-execute-list/v4.ts
vikinatora: I'll talk to the PO about this

vikinatora: If we allow `reservoir`, the fees that our contract honours and that are set in the order itself will be persisted right?

package.json
```diff
@@ -33,7 +33,7 @@
     "@hapi/inert": "^6.0.4",
     "@hapi/vision": "^6.1.0",
     "@poprank/rankings": "^1.1.20",
-    "@reservoir0x/sdk": "0.0.182",
```
georgeroman, line 36: Let's `yarn add @reservoir0x/sdk@0.0.183` so that `yarn.lock` gets updated as well.

PR #1991, opened 2022-09-28
Fix Universe order creating + filling logic

vikinatora:
Changes:
1. Various validation + execution logic issues that I encountered during creating and filling sell orders from our FE dev environment.
2. Added missing logic inside `post-order/v3.ts`.
3. Salt calculation that's based on the amount of orders the user has created. This ensures that hashOrderKey() will always return a different hash even if all other order parameters are the same

To be added:
1. SDK package version to be `0.0.184` after https://github.com/reservoirprotocol/core/pull/36 is merged

### DIFF
--- a/src/orderbook/orders/universe/build/utils.ts
+++ b/src/orderbook/orders/universe/build/utils.ts
@@ -2,7 +2,6 @@ import * as Sdk from "@reservoir0x/sdk";
 
 import { redb } from "@/common/db";
 import { OrderSide } from "@reservoir0x/sdk/dist/universe/types";
-import { toBuffer } from "@/common/utils";
 
 export interface BaseOrderBuildOptions {
   maker: string;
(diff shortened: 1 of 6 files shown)


src/orderbook/orders/universe/build/utils.ts
```diff
@@ -49,20 +50,30 @@ export const getBuildInfo = async (
     throw new Error("Invalid NFT asset class");
   }
 
+  const makerOrdersCount = await redb.many(
```
georgeroman, line 53: This is not going to scale on production where we have millions of orders. Is it that important to have the salts increasing?

vikinatora:
```
  public hashOrderKey() {
    const encodedOrder = utils.defaultAbiCoder.encode(
      ["address", "bytes32", "bytes32", "uint256"],
      [
        lc(this.params.maker),
        hashAssetType(this.params.make.assetType),
        hashAssetType(this.params.take.assetType),
        Number(this.params.salt),
      ]
    );

    return utils.keccak256(encodedOrder);
  }
```

It's important to have them not repeated for the same user. This ensures that we won't have the same hash if a user creates the exact same order twice(for whatever reason). If we don't have unique hashes the contract won't work as expected. 

I'm open to other suggestions for salt generation. This is the one we used on our backend

PR #2142, opened 2022-10-25
feat: support element

lljxx1:
- add element
- `bulkCancelEvents` add support to cancle all the maker's related orders
- integration tests cover `sellERC721` `buyERC721` `sellERC1155` `buyERC1155` `cancelOrder` `incrementHashNonce`


src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,455 @@
...
+      // Handle: get order kind
+      const kind = await commonHelpers.getContractKind(order.params.nft);
+      if (!kind) {
+        return results.push({
+          id,
+          status: "unknown-order-kind",
+        });
+      }
+
+      // Check: order has unique nonce
```
georgeroman, line 72: Since the Element filling events include the order's hash there is no need to have this restriction here. This is only relevant for ZeroExV4 which doesn't include the order's hash in the fill event and so we had to have some uniqueness restrictions in order to know which order to mark as filled.

src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,455 @@
...
+              schemaHash,
+              contract: order.params.nft,
+              tokenId: order.params.nftId,
+            },
+          ]);
+
+          break;
+        }
+
+        case "token-range": {
```
georgeroman, line 237: No need to support `token-range` and `token-list-*`, since Element only supports `single-token` and `contract-wide`.

src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,455 @@
...
+        return results.push({
+          id,
+          status: "invalid-token-set",
+        });
+      }
+
+      // Handle: fees
+      const feeAmount = order.getFeeAmount();
+
+      // const timestamp = BigNumber.from(order.params.expiry).and(BigNumber.from("0xffffffff")).toString();
```
georgeroman, line 292: Should remove this

src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,455 @@
...
+        token_set_schema_hash: toBuffer(schemaHash),
+        maker: toBuffer(order.params.maker),
+        taker: toBuffer(order.params.taker),
+        price: price.toString(),
+        value: value.toString(),
+        currency: toBuffer(currency),
+        currency_price: price.toString(),
+        currency_value: value.toString(),
+        needs_conversion: null,
+        quantity_remaining: order.params.nftAmount,
```
georgeroman, line 359: Will this work for ERC721 as well? We want a default of 1 in this case.

src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,455 @@
...
+              context: `new-order-${id}`,
+              id,
+              trigger: {
+                kind: "new-order",
+              },
+            } as ordersUpdateById.OrderInfo)
+        )
+    );
+
+    // if (relayToArweave) {
```
georgeroman, line 449: Should enable this?

georgeroman (changes requested):
We don't need to create Element orders, so let's remove any logic in `/execute/bid`, `/execute/list` and `POST /order` (also in `/orders/element/build`). We should probably support element in `POST /orders/` (since that's what the relayer uses).

src/sync/events/storage/bulk-cancel-events.ts
```diff
@@ -25,10 +26,61 @@ type DbEvent = {
...
+      RETURNING "order_kind", "maker", "min_nonce", "tx_hash", "timestamp", "log_index", "batch_index", "block_hash"
+    )
+    UPDATE "orders" AS "o" SET
+      "fillability_status" = 'cancelled',
+      "expiration" = to_timestamp("x"."timestamp"),
+      "updated_at" = now()
+    FROM "x"
+    WHERE "o"."maker" = "x"."maker"
+      And "o"."kind" = "x"."order_kind"
+      ${acrossAll ? `` : `AND "o"."nonce" < "x"."min_nonce" `}
```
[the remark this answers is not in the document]
lljxx1: yeah, currently there has two kind of nonce: 
- order-level 
- account-level 

the order's nonce in database is order-level.
when account-level nonce increase, cannot use `nonce < min nonce` to cancel all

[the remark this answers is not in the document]
lljxx1: @georgeroman That's why I need to remove the `nonce < main nonce` condition

src/api/endpoints/orders/post-order/v2.ts
```diff
@@ -11,7 +11,7 @@ import { config } from "@/config/index";
 import * as orders from "@/orderbook/orders";
 
 import * as postOrderExternal from "@/jobs/orderbook/post-order-external";
-
+f;
```
georgeroman, line 14: Let's remove this and should be good to go on dev

src/orderbook/orders/element/index.ts
```diff
@@ -0,0 +1,354 @@
...
+          status: "fees-too-high",
+        });
+      }
+
+      // Handle: source
+      const sources = await Sources.getInstance();
+      const source = metadata.source ? await sources.getOrInsert(metadata.source) : undefined;
+
+      // Handle: native Reservoir orders
+      const isReservoir = true;
```
georgeroman, line 224: This should be false.

PR #2179, opened 2022-11-02
Feat/whitelisted currencies

nofir:


src/utils/prices/index.ts
```diff
@@ -72,6 +74,34 @@ const getUpstreamUSDPrice = async (
           value,
         };
       }
+    } else if (getNetworkSettings().whitelistedCurrencies.has(currencyAddress)) {
+      const usdPrice = 0.000001;
+      const value = parseUnits(usdPrice.toFixed(USD_DECIMALS), USD_DECIMALS).toString();
```
georgeroman, line 79: Could we simplify this logic to `const value = "1"`? Are all these currencies 1:1 with USD?

nofir: yeah, makes sense.

PR #2190, opened 2022-11-03
Ofir/res 949 migrate to use opensea stream api for

nofir:


src/websockets/opensea/index.ts
```diff
@@ -2,19 +2,76 @@ import { Network, OpenSeaStreamClient } from "@opensea/stream-js";
...
+          kind: "partial",
+          orderParams: {
+            kind: "single-token",
+            side: "sell",
+            hash: event.payload.order_hash,
+            price: event.payload.base_price,
+            paymentToken: event.payload.payment_token.address,
+            amount: event.payload.quantity,
+            startTime: toTime(event.payload.listing_date),
+            endTime: toTime(event.payload.listing_date),
```
georgeroman, line 46: Do they have the `expiration_date`?

nofir: they do, i just made a change and copy pasted the startTime line... thanks for noticing. fixed.

PR #2231, opened 2022-11-08
Jadendurnford/res 998 return quantityfilled and quantityremaining in tokens v5

JadenDurnford:


src/api/endpoints/tokens/get-tokens/v5.ts
```diff
@@ -343,7 +345,21 @@ export const getTokensV5Options: RouteOptions = {
               AND nb.token_id = t.token_id
               AND nb.amount > 0
             LIMIT 1
-          ) AS owner
+          ) AS owner,
+          (
```
georgeroman, line 349: Take a look at `JOIN LATERAL` so that we can avoid two `SELECT`s on the `orders` table.

georgeroman: Also, the way we work with the branches is the following:
- every new branch should be branched off `v5` (the prod branch -> deployments going to mainnets)
- the first PR should be against `v5-dev` (the dev branch -> deployments going to testnets)
- once everything is tested and works fine on dev we create a PR against `v5` and merge to prod

ipeleg: Small correction on the first bullet point, you should branch off `main` and not `v5`

georgeroman:
LGTM! Let's first merge to v5-dev though, for proper testing.

PR #2293, opened 2022-11-15
feat: add blur

lljxx1:


src/api/endpoints/orders/post-order/v3.ts
```diff
@@ -43,7 +43,9 @@ export const postOrderV3Options: RouteOptions = {
             "seaport-forward",
             "x2y2",
             "universe",
-            "forward"
+            "forward",
+            "element",
```
georgeroman, line 47: I don't think we need these here, Element and Blur orders are only going to be sent by the relayer to the `/orders` API while this API is designed for pushing orders created via the execute APIs.

src/orderbook/orders/blur/index.ts
```diff
@@ -0,0 +1,304 @@
...
+        // Keep any orders that can potentially get valid in the future
+        if (error.message === "no-balance-no-approval") {
+          fillabilityStatus = "no-balance";
+          approvalStatus = "no-approval";
+        } else if (error.message === "no-approval") {
+          approvalStatus = "no-approval";
+        } else if (error.message === "no-balance") {
+          fillabilityStatus = "no-balance";
+        } else {
+          //   return results.push({
```
georgeroman, line 136: This shouldn't be commented out, otherwise we'll ingest invalid orders.

src/orderbook/orders/blur/index.ts
```diff
@@ -0,0 +1,304 @@
...
+          id,
+          status: "invalid-token-set",
+        });
+      }
+
+      // Handle: fees
+      const side = order.params.side === Sdk.Blur.Types.TradeDirection.BUY ? "buy" : "sell";
+
+      // Handle: price and value
+      const price = bn(order.params.price);
```
georgeroman, line 166: Is the amount always 1? Also, are Blur orders partially fillable? If not, then we should make sure to scale the price by the amount - the price in the orders table should be the price of a single unit.

src/orderbook/orders/index.ts
```diff
@@ -49,8 +50,7 @@ export type OrderKind =
   | "universe"
   | "nftx"
   | "blur"
-  | "forward"
-  | "rarible";
```
georgeroman, line 53: Let's keep `rarible` here.

georgeroman (changes requested):

src/api/endpoints/orders/post-order/v3.ts
georgeroman: So let's make sure to add support for Element and Blur in the `/orders` API.

src/orderbook/orders/index.ts
lljxx1: `rarible` in the top lol https://github.com/reservoirprotocol/indexer/blob/62d596854c489ae369231ae10672d21c940392bf/src/orderbook/orders/index.ts#L41

src/orderbook/orders/blur/index.ts
lljxx1: https://github.com/lljxx1/nft-exchanges/blob/59757953527342c2c0f1787c5b10a53eb6d03537/Blur/contracts/BlurExchange.sol#L425
https://github.com/lljxx1/nft-exchanges/blob/main/Blur/contracts/matchingPolicies/StandardPolicyERC721.sol#L33

Check the code. For now, It's always 1

src/api/endpoints/orders/post-order/v3.ts
lljxx1: 👌

src/orderbook/orders/index.ts
```diff
@@ -37,7 +38,6 @@ export type OrderKind =
   | "foundation"
   | "x2y2"
   | "seaport"
-  | "rarible"
```
georgeroman, line 40: Let's add `rarible` back in.

PR #2390, opened 2022-11-22
fix: shared storefront tokens contract mixup

JadenDurnford:


JadenDurnford:
removed extra commits which were included on the branch

src/sync/events/handlers/seaport.ts
```diff
@@ -121,6 +121,10 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
           }
 
           const orderSide = saleInfo.side as "sell" | "buy";
+          let contract = saleInfo.contract;
```
georgeroman, line 124: Maybe let's just add a comment about the reasoning behind this mapping

PR #2392, opened 2022-11-22
Jaden/res 1127 when normalizedroyalties is set to true

JadenDurnford:


src/api/endpoints/orders/get-orders-bids/v4.ts
```diff
@@ -468,6 +469,24 @@ export const getOrdersBidsV4Options: RouteOptions = {
 
       const sources = await Sources.getInstance();
       const result = rawResult.map(async (r) => {
+        const feeBreakdown = r.fee_breakdown;
+        let feeBps = Number(r.fee_bps);
+
+        if (query.normalizeRoyalties) {
+          for (let i = 0; i < r.missing_royalties.length; i++) {
+            const bps =
+              (r.missing_royalties[i].amount /
```
nofir, line 478: i would cast the amount to big number when doing this calculation

nofir (changes requested):

PR #2433, opened 2022-11-24
feat: simulate collection floor orders to ensure they are fillable

JadenDurnford:


ptrwtts:
Hey Jaden,

This code will prevent the collection floor from being updated if the token is either flagged or the simulation fails. We always want to update the collection floor as we currently do. 

Instead, what we want is to do some extra checks AFTER updating the floor, in the event that the collection floor changed. So this logic should happen within the `if (collectionFloorAskChanged) {` and the checks we want to trigger are:

- Recheck if the token is flagged (just so we have the latest info. It's still allowed to be floor if flagged)
- Simulate if the order is fillable (if it's not, it will automatically be cleaned up as part of simulation, which will trigger a new floor update)

These extra checks can happen asynchronously. It shouldn't impact the regular flow of updating the collection floor.

nofir:
created a new branch with the correct implementation, closing this pr

PR #2438, opened 2022-11-28
Manifold

boris-bonin:
Initial indexer implementation that support cancel, purchase, finalize and modify  event ingestion.

Manifold is an on-chain escrowed orderbook.

### DIFF
--- a/src/jobs/orderbook/orders-queue.ts
+++ b/src/jobs/orderbook/orders-queue.ts
@@ -98,7 +98,8 @@ if (config.doBackgroundWork) {
           }
 
           case "rarible": {
-            result = await orders.rarible.save([info], relayToArweave);
+            const result = await orders.rarible.save([info], relayToArweave);
+            logger.info(QUEUE_NAME, `[rarible] Order save result: ${JSON.stringify(result)}`);
             break;
           }
 
@@ -113,7 +114,8 @@ if (config.doBackgroundWork) {
           }
 
           case "manifold": {
-            result = await orders.manifold.save([info]);
+            const result = await orders.manifold.save([info]);
+            logger.info(QUEUE_NAME, `[manifold] Order save result: ${JSON.stringify(result)}`);
             break;
           }
         }
--- a/src/jobs/orderbook/orders-queue.ts
+++ b/src/jobs/orderbook/orders-queue.ts
@@ -98,9 +98,7 @@ if (config.doBackgroundWork) {
           }
 
           case "rarible": {
-            const result = await orders.rarible.save([info], relayToArweave);
-            logger.info(QUEUE_NAME, `[rarible] Order save result: ${JSON.stringify(result)}`);
-
+            result = await orders.rarible.save([info], relayToArweave);
             break;
           }
 
@@ -115,9 +113,7 @@ if (config.doBackgroundWork) {
           }
 
           case "manifold": {
-            const result = await orders.manifold.save([info]);
-            logger.info(QUEUE_NAME, `[manifold] Order save result: ${JSON.stringify(result)}`);
-
+            result = await orders.manifold.save([info]);
             break;
           }
         }
--- a/src/jobs/orderbook/orders-queue.ts
+++ b/src/jobs/orderbook/orders-queue.ts
@@ -115,7 +115,7 @@ if (config.doBackgroundWork) {
           }
 
           case "manifold": {
-            const result = await orders.manifold.save([info], relayToArweave);
+            const result = await orders.manifold.save([info]);
             logger.info(QUEUE_NAME, `[manifold] Order save result: ${JSON.stringify(result)}`);
 
             break;
--- a/src/jobs/orderbook/orders-queue.ts
+++ b/src/jobs/orderbook/orders-queue.ts
@@ -113,6 +113,13 @@ if (config.doBackgroundWork) {
             result = await orders.blur.save([info as orders.blur.OrderInfo], relayToArweave);
             break;
           }
+
+          case "manifold": {
+            const result = await orders.manifold.save([info], relayToArweave);
+            logger.info(QUEUE_NAME, `[manifold] Order save result: ${JSON.stringify(result)}`);
+
+            break;
+          }
         }
       } catch (error) {
         logger.error(QUEUE_NAME, `Failed to process order ${JSON.stringify(job.data)}: ${error}`);
@@ -226,6 +233,12 @@ export type GenericOrderInfo =
   | {
       kind: "blur";
       info: orders.blur.OrderInfo;
+      relayToArweave?: boolean;
+      validateBidValue?: boolean;
+    }
+  | {
+      kind: "manifold";
+      info: orders.manifold.OrderInfo;
       relayToArweave?: boolean;
       validateBidValue?: boolean;
     };
--- a/src/orderbook/orders/manifold/check.ts
+++ b/src/orderbook/orders/manifold/check.ts
@@ -4,7 +4,6 @@ import { baseProvider } from "@/common/provider";
 import { config } from "@/config/index";
 import { OrderInfo } from "@/orderbook/orders/manifold";
 import { bn } from "@/common/utils";
-import { constants } from "ethers";
 
 export const offChainCheck = async (order: OrderInfo["orderParams"]) => {
   const exchangeContract = new Sdk.Manifold.Exchange(config.chainId);
@@ -43,18 +42,5 @@ export const offChainCheck = async (order: OrderInfo["orderParams"]) => {
 
   if (onChainListing.details.totalPerSale != order.details.totalPerSale) {
     throw Error("invalid");
-  }
-
-  // Indexer specififc validations
-  if (order.details.totalPerSale !== 1) {
-    throw Error("unfillable");
-  }
-
-  if (order.details.type_ !== 2) {
-    throw Error("unfillable");
-  }
-
-  if (order.details.erc20 !== constants.AddressZero) {
-    throw Error("unfillable");
   }
 };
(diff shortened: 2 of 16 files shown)


vikinatora:
The only sell events that we're not indexing right now are ranked auction and dynamic price listing. The issue with them are that token transfer and nft delivery is done in collect and deliver function. These functions don't emit any exchange specific events and it's hard to track. This is not an issue in my opinion as there's not been such orders created on Manifold. Check https://marketplace.api.manifoldxyz.dev/listing/0x3a3548e060be10c2614d0a4cb0c03cc9093fd799/activity?sortAscending=true&pageNumber=1&pageSize=2000&type=2 the two order types that we don't support are `type_: 3` and `type_: 4`.

src/sync/events/handlers/manifold.ts
```diff
@@ -0,0 +1,460 @@
...
+import * as utils from "@/events-sync/utils";
+
+import * as fillUpdates from "@/jobs/fill-updates/queue";
+import * as orderUpdatesById from "@/jobs/order-updates/by-id-queue";
+import * as orderUpdatesByMaker from "@/jobs/order-updates/by-maker-queue";
+import { getOrderId, OrderInfo } from "@/orderbook/orders/manifold";
+import { manifold } from "@/orderbook/orders";
+import { getUSDAndNativePrices } from "@/utils/prices";
+import { CallTrace } from "@georgeroman/evm-tx-simulator/dist/types";
+import { BigNumber } from "ethers";
```
georgeroman, line 18: Let's use the `bn` method instead.

src/sync/events/handlers/manifold.ts
```diff
@@ -0,0 +1,460 @@
...
+type TransferEventWithContract = es.nftTransfers.Event & { tokenContract: string };
+
+const findEthCalls = (calls: CallTrace[], eventParams: BaseEventParams) => {
+  if (!calls || !calls.length) {
+    return [];
+  }
+
+  const ethCalls: CallTrace[] = [];
+  calls.forEach((call: CallTrace) => {
+    if (call.type === "CALL" && call.from === eventParams.address && call.value !== "0x0") {
```
georgeroman, line 30: Hmm, what's the idea behind the `call.from === eventParams.address` check?

src/sync/events/handlers/manifold.ts
```diff
@@ -0,0 +1,460 @@
...
+            // Maker is the receiver of tokens
+            maker = to;
+            currency = log.address.toLowerCase();
+            currencyPrice = amount;
+            break;
+
+            // If we don't detect an ERC20 transfer as part of the same transaction
+            // then we assume it's ETH
+          } else {
+            // Event data doesn't include full transaction information so we have to parse the calldata
```
georgeroman, line 313: Same here, better if we could use call tracing for everything instead.

georgeroman (changes requested):

src/sync/events/handlers/manifold.ts
vikinatora: Maybe `call.from === eventParams.address` isn't necessary but since ETH transfer always have the exchange contract as `from` I decided to add that as additional condition

src/sync/events/handlers/manifold.ts
```diff
@@ -0,0 +1,460 @@
...
+            log.topics[0] === erc20EventData.topic &&
+            log.topics.length === erc20EventData.numTopics
+          ) {
+            const parsed = erc20EventData.abi.parseLog(log);
+            const to = parsed.args["to"].toLowerCase();
+            const amount = parsed.args["amount"].toString();
+            // Maker is the receiver of tokens
+            maker = to;
+            currency = log.address.toLowerCase();
+            currencyPrice = amount;
```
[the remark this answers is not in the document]
vikinatora: We do that here: https://github.com/reservoirprotocol/indexer/pull/2438/files#diff-3d80e2d160235555d4301cf1d94e6249ff4b3fc9442ec55720808dd58e8a5f2aR342

src/jobs/orderbook/orders-queue.ts
```diff
@@ -98,9 +98,7 @@ if (config.doBackgroundWork) {
           }
 
           case "rarible": {
-            const result = await orders.rarible.save([info], relayToArweave);
-            logger.info(QUEUE_NAME, `[rarible] Order save result: ${JSON.stringify(result)}`);
```
georgeroman, line 102: Let's leave the log, and also include it for Manifold orders.

src/orderbook/orders/manifold/index.ts
```diff
@@ -0,0 +1,301 @@
...
+        fillability_status: fillabilityStatus,
+        approval_status: approvalStatus,
+        token_set_id: tokenSetId,
+        token_set_schema_hash: toBuffer(schemaHash),
+        maker: toBuffer(orderParams.seller),
+        taker: toBuffer(AddressZero),
+        price: orderParams.details.initialAmount,
+        value: orderParams.details.initialAmount,
+        // erc20 is zero address if ETH order
+        currency: toBuffer(orderParams.details.erc20),
```
georgeroman, line 209: Let's just ignore any non-ETH orders.

src/orderbook/orders/manifold/index.ts
```diff
@@ -0,0 +1,301 @@
...
+        currency_price: orderParams.details.initialAmount,
+        currency_value: orderParams.details.initialAmount,
+        needs_conversion: null,
+        quantity_remaining: orderParams.details.totalAvailable.toString(),
+        valid_between: `tstzrange(${validFrom}, ${validTo}, '[]')`,
+        nonce: null,
+        source_id_int: source?.id,
+        is_reservoir: null,
+        contract: toBuffer(contract),
+        conduit: toBuffer(Sdk.Manifold.Addresses.Exchange[config.chainId]),
```
georgeroman, line 219: Let's set the conduit to `null` since it's not relevant for escrowed orderbooks.

src/sync/events/handlers/manifold.ts
```diff
@@ -0,0 +1,348 @@
...
+              raw_data,
+              extract('epoch' from lower(orders.valid_between)) AS valid_from
+            FROM orders 
+            WHERE orders.id = $/id/ 
+          `,
+          { id: orderId }
+        );
+
+        if (!orderResult) {
+          break;
```
georgeroman, line 94: If we're skipping sales for which we don't have the order it means that we won't have historical sales, which we want. Any fields extracted from the order data can probably be retrieved from the call trace.

georgeroman (changes requested):

src/orderbook/orders/manifold/index.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+
+export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
+  const results: SaveResult[] = [];
+  const orderValues: DbOrder[] = [];
+
+  const handleOrder = async ({ orderParams, metadata }: OrderInfo) => {
+    try {
+      const id = getOrderId(orderParams.id);
+
```
georgeroman, line 48: Should we also ensure here the listing is a fixed-price one (eg. type = 2)?

georgeroman (changes requested):
One additional change we might want to make is make sure to trigger an `orderUpdatesById` for in-place order updates (and not just for new orders). Check out these changes: https://github.com/reservoirprotocol/indexer/pull/2513/files.

src/orderbook/orders/manifold/index.ts
```diff
@@ -0,0 +1,304 @@
...
+        "raw_data",
+        { name: "expiration", mod: ":raw" },
+      ],
+      {
+        table: "orders",
+      }
+    );
+    await idb.none(pgp.helpers.insert(orderValues, columns) + " ON CONFLICT DO NOTHING");
+
+    await ordersUpdateById.addToQueue(
```
georgeroman, line 287: We should move this outside of the if statement, otherwise we'll only trigger events if the current batch contains new orders (which might not be the case at all times).

src/orderbook/orders/manifold/check.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  if (onChainListing.details.totalAvailable != order.details.totalAvailable) {
+    throw Error("invalid");
+  }
+
+  if (onChainListing.details.totalPerSale != order.details.totalPerSale) {
+    throw Error("invalid");
+  }
+
+  // Indexer specififc validations
```
georgeroman, line 48: Indexer-specific validations should go into the `index.ts` file (in the `save` method).

src/orderbook/orders/manifold/check.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (order.details.totalPerSale !== 1) {
+    throw Error("unfillable");
+  }
+
+  if (order.details.type_ !== 2) {
+    throw Error("unfillable");
+  }
+
+  if (order.details.erc20 !== constants.AddressZero) {
```
georgeroman, line 57: Should use `Sdk.Common.Addresses.Eth` instead of `AddressZero`.

src/sync/events/handlers/manifold.ts
georgeroman: I'm worried about historical listings which might not get ingested if we have this check. Maybe we can assume that the maker is the recipient of the highest ETH/ERC20 payment part of the purchase (sub)transaction. What do you think?

src/sync/events/handlers/manifold.ts
```diff
@@ -158,8 +166,19 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
+              extract('epoch' from lower(orders.valid_between)) AS valid_from
+            FROM orders 
+            WHERE orders.id = $/id/ 
+          `,
+          { id: orderId }
+        );
+
         // Some manifold order have end time that is set after the first purchase
-        if (orderResult.valid_from === 0) {
+        if (orderResult && orderResult.valid_from === 0) {
```
vikinatora, line 181: This should stay as it's applicable only to active orders.

PR #2470, opened 2022-11-29
Infinity

jfrazier-eth:
Adds support for Infinity marketplace
Expected Support: 
- [X]  Handle Infinity contract events - cancels/sales
- [X] Store orders and maintain current status
- [X] Support filling orders natively 
- [X] Support for publishing orders to the Infinity orderbook

Future work: 
* Support m of n order validity/fillability 
  * Requires multi-collection token sets
* Add support for scraping orders to the relayer - I'm planning on implementing some new order endpoints in the next week or two and will update the relayer once that is complete


src/orderbook/orders/index.ts
```diff
@@ -106,6 +108,8 @@ export const getOrderSourceByOrderKind = async (
         return sources.getOrInsert("nftx.io");
       case "blur":
         return sources.getOrInsert("blur.io");
+      case "infinity":
+        return sources.getOrInsert("infinity.xyz"); // TODO - what needs to be done to add infinity to source? do I update the json file?
```
georgeroman, line 112: No need to do anything, we'll fetch all details from the domain automatically. Can manually tweak anything afterwards. Let's remove the comment.

src/orderbook/orders/infinity/build/buy/collection.ts
```diff
@@ -0,0 +1,36 @@
+import * as Sdk from "@reservoir0x/sdk";
+
+import { redb } from "@/common/db";
+import { config } from "@/config/index";
+import * as utils from "../utils";
```
georgeroman, line 5: Not a big deal, but we usually use full path imports: `@/orderbook/orders/infinity/build/utils`.

src/orderbook/orders/infinity/build/buy/token.ts
```diff
@@ -0,0 +1,42 @@
+import * as Sdk from "@reservoir0x/sdk";
+
+import { redb } from "@/common/db";
+import { config } from "@/config/index";
+import * as utils from "../utils";
```
georgeroman, line 5: Same comment as above for the full path import.

src/orderbook/orders/infinity/build/sell/token.ts
```diff
@@ -0,0 +1,42 @@
+import * as Sdk from "@reservoir0x/sdk";
+
+import { redb } from "@/common/db";
+import { config } from "@/config/index";
+import * as utils from "../utils";
```
georgeroman, line 5: Same here.

src/orderbook/orders/infinity/build/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+export interface BaseOrderBuildOptions {
+  maker: string;
+  weiPrice: string;
+  currency?: string;
+  nonce?: string;
+  listingTime?: number;
+  expirationTime?: number;
+  maxGasPrice?: number;
```
georgeroman, line 21: Interesting, is this an order parameter?

src/orderbook/orders/infinity/check.ts
```diff
@@ -0,0 +1,124 @@
...
+    // of buy orders as well.
+    onChainApprovalRecheck?: boolean;
+    checkFilledOrCancelled?: boolean;
+  }
+) => {
+  const contracts = order.nfts.map((item) => item.collection);
+
+  for (const contract of contracts) {
+    const kind = await commonHelpers.getContractKind(contract);
+    if (!kind) {
```
georgeroman, line 28: Simpler to use `kind !== "erc721"`?

src/orderbook/orders/infinity/index.ts
```diff
@@ -0,0 +1,328 @@
...
+      const currentTime = now();
+      const orderExpires = order.endTime !== 0;
+      // Check: order is not expired
+      if (orderExpires && currentTime >= order.endTime) {
+        return results.push({
+          id,
+          status: "expired",
+        });
+      }
+
```
georgeroman, line 64: Let's maybe add a check here to ensure the order only applies to a single NFT, eg. `order.nfts.length === 1`.

src/sync/events/handlers/infinity.ts
```diff
@@ -0,0 +1,438 @@
...
+          currency,
+          pricePerToken,
+          baseEventParams.timestamp
+        );
+        if (!priceDataPerToken.nativePrice) {
+          // We must always have the native price
+          break;
+        }
+
+        for (const nft of nfts) {
```
georgeroman, line 141: Are orders with multiple tokens bundles? If so we should probably exclude them for the fill events (at least for now) since it doesn't make sense to split the price across all tokens in a bundle.

src/orderbook/orders/infinity/index.ts
georgeroman, line 64: We should also make sure to only support ETH listings and WETH bids.

src/sync/events/handlers/infinity.ts
```diff
@@ -0,0 +1,438 @@
...
+
+  if (result?.nonce) {
+    return (result.nonce as BigNumberish).toString();
+  }
+
+  return null;
+}
+
+async function getOrderSide(orderHash: string, seller: string, buyer: string) {
+  const sellerOrderNonce = await getOrderNonce(orderHash, seller);
```
georgeroman, line 417: What would happen if we don't have the order in our system so we can't cancel the nonce but then we receive an order with the nonce that was cancelled via filling? We should probably use call tracing (check the sudoswap / blur sales handling) to always get the nonce (I assume it's available in the calldata).

georgeroman (changes requested):
Let's remove `package-lock.json`

src/orderbook/orders/infinity/index.ts
jfrazier-eth: What is the reasoning for adding this here? To prevent the `offChainCheck` from happening?

src/orderbook/orders/infinity/build/utils.ts
jfrazier-eth: Yes, when orders are executed by our matching engine our exchange takes refund from the buyer to pay for any gas used. The max gas price can be set so that they can control the max gas price that their order will be executed at.

src/orderbook/orders/infinity/build/buy/collection.ts
jfrazier-eth: Will do, would you like me to add a `.vscode/settings.json` with the setting below to make this the default for vscode auto imports?
```
{
  "typescript.preferences.importModuleSpecifier": "non-relative"
}
```

src/orderbook/orders/index.ts
jfrazier-eth: Great

src/sync/events/handlers/infinity.ts
jfrazier-eth: Okay, does this include fill info events? i.e. should we exclude both fill events and fill info events for bundles?

src/orderbook/orders/infinity/index.ts
georgeroman: Yes, just to return as early as possible.

src/sync/events/handlers/infinity.ts
georgeroman: Yes, since it can be misleading to split the price across all items in the bundle, eg. a bundle of one bored ape and one worthless token selling at 100ETH would split the price across the two, when in reality one token was more expensive than the other.

src/sync/events/handlers/infinity.ts
jfrazier-eth: Good catch, I added parsing the nonce from the calldata

src/orderbook/orders/infinity/index.ts
```diff
@@ -0,0 +1,353 @@
...
+        currency_price: price.toString(),
+        currency_value: value.toString(),
+        needs_conversion: null,
+        quantity_remaining: order.numItems.toString(),
+        valid_between: `tstzrange(${validFrom}, ${validTo}, '[]')`,
+        nonce: order.nonce,
+        source_id_int: source?.id,
+        is_reservoir: isReservoir,
+        contract: null,
+        conduit: null,
```
georgeroman, line 245: The conduit should be the contract on which approvals are given, required because that's how we are going to validate/invalidate orders given approval changes.

georgeroman (changes requested):

src/orderbook/orders/infinity/index.ts
```diff
@@ -0,0 +1,353 @@
...
+        currency: toBuffer(order.currency),
+        currency_price: price.toString(),
+        currency_value: value.toString(),
+        needs_conversion: null,
+        quantity_remaining: order.numItems.toString(),
+        valid_between: `tstzrange(${validFrom}, ${validTo}, '[]')`,
+        nonce: order.nonce,
+        source_id_int: source?.id,
+        is_reservoir: isReservoir,
+        contract: null,
```
georgeroman, line 244: The contract should also be set to the contract the order applies to (we don't support any type of cross-contract orders at the moment).

src/orderbook/orders/infinity/index.ts
```diff
@@ -241,8 +241,8 @@ export const save = async (orderInfos: OrderInfo[], relayToArweave?: boolean) =>
         nonce: order.nonce,
         source_id_int: source?.id,
         is_reservoir: isReservoir,
-        contract: null,
-        conduit: null,
+        contract: toBuffer(Sdk.Infinity.Addresses.Exchange[config.chainId]),
```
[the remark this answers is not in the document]
jfrazier-eth: Ahh, will change this

PR #2480, opened 2022-12-01
Feat: consecutive transfers

vikinatora:
Added event hander for erc721 consescutive transfers. The processing is almost identical to erc721 single transfers. The event handler logic for both events can be extracted in a function but decided to not go for that since their handling should be independent of one another.

### DIFF
--- a/src/sync/events/handlers/erc721.ts
+++ b/src/sync/events/handlers/erc721.ts
@@ -75,6 +75,21 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
           },
         });
 
+        makerInfos.push({
+          context: `${contextPrefix}-${to}-sell-balance`,
+          maker: to,
+          trigger: {
+            kind: "balance-change",
+            txHash: baseEventParams.txHash,
+            txTimestamp: baseEventParams.timestamp,
+          },
+          data: {
+            kind: "sell-balance",
+            contract: baseEventParams.address,
+            tokenId,
+          },
+        });
+
         if (from === AddressZero) {
           mintInfos.push({
             contract: baseEventParams.address,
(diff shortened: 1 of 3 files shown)


src/sync/events/handlers/erc721.ts
```diff
@@ -75,21 +75,6 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
           },
         });
 
-        makerInfos.push({
-          context: `${contextPrefix}-${to}-sell-balance`,
```
georgeroman, line 79: Just realized that this is actually not a duplicate, it has a different context (using `to` instead of `from`).

src/sync/events/handlers/erc721.ts
```diff
@@ -114,6 +99,68 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
+                baseEventParams,
+              });
+            }
+          }
+
+          // Make sure to only handle the same data once per transaction
+          const contextPrefix = `${baseEventParams.txHash}-${baseEventParams.address}-${tokenId}`;
+
+          makerInfos.push({
+            context: `${contextPrefix}-${from}-sell-balance`,
```
georgeroman, line 146: We need a similar `makerInfo` struct for the `to` address, as mentioned above regarding the duplicate.

georgeroman (changes requested):

src/sync/events/handlers/erc721.ts
```diff
@@ -114,6 +99,68 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
       }
 
+      case "erc721-consecutive-transfer": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const from = parsedLog.args["fromAddress"].toLowerCase();
+        const to = parsedLog.args["toAddress"].toLowerCase();
+        const fromTokenId = parsedLog.args["fromTokenId"].toString();
+        const toTokenId = parsedLog.args["toTokenId"].toString();
+
+        for (let i = fromTokenId; i <= toTokenId; i++) {
```
georgeroman, line 109: Since `fromTokenId` and `toTokenId` are `string`s, incrementing will not work I think. What you could do instead is wrap the `for` loop in a `try`/`catch` and convert to `number`.

georgeroman (changes requested):

src/sync/events/handlers/erc721.ts
vikinatora: Converting strings works implicitly. I've tested it in multiple scenarios. However I agree that the iterators should work with `number` instead of `string`

vikinatora:
@georgeroman Ready for re-review

PR #2498, opened 2022-12-02
Cryptokitties transfer handling

vikinatora:
Added handling of crypto kitties. We need to get the SDK PR merged first in order to upgrade the package version ehre


vikinatora:
@georgeroman This can be reviewed as well

PR #2551, opened 2022-12-07
Non standard tokens - Beeple, CryptoArte, Cryptovoxels transfer handling

vikinatora:
The transfer events look identical for these two contracts. Not sure if we want to abstract these events in some way.
SDK PR: https://github.com/reservoirprotocol/core/pull/62


src/sync/events/storage/nft-transfer-events.ts
```diff
@@ -7,7 +7,14 @@ import _ from "lodash";
 import { logger } from "@/common/logger";
 
 export type Event = {
-  kind: "erc721" | "erc1155" | "cryptopunks" | "cryptokitties";
+  kind:
```
georgeroman, line 10: Let's extract these to a high-level type `ContractKind`.

src/migrations/1670318026167_add-cryptoarte-contract-kind.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+
+ALTER TYPE "contract_kind_t" ADD VALUE 'cryptoarte';
```
georgeroman, line 3: Since all of these (including the previously added cryptokitties) have the same event signature I think we can bundle all together into a single contract kind `erc20-like` (since the event signature matches the ERC20 transfer signature).

georgeroman: This should result in a single handler for all of these.

georgeroman: So let's just rename `cryptokitties` to `erc20-like` (eg. https://stackoverflow.com/questions/12627830/rename-enum-item-in-postgresql).

src/sync/events/handlers/index.ts
```diff
@@ -162,6 +168,21 @@ export const processEvents = async (info: EventsInfo) => {
       data = await manifold.handleEvents(info.events);
       break;
     }
+
+    case "cryptoarte": {
```
georgeroman, line 172: Once we have the `erc20-like` type, we should use the same handler as we do for `erc721`: https://github.com/reservoirprotocol/indexer/blob/cb747506c692db144323eab6824a4c7296c8e15a/src/sync/events/handlers/index.ts#L62
We should only make sure to handle it in the exact way as a `erc721-transfer`: https://github.com/reservoirprotocol/indexer/blob/cb747506c692db144323eab6824a4c7296c8e15a/src/sync/events/handlers/erc721.ts#L45

Something like:
```javascript
case "erc721-transfer":
case "erc20-like-transfer": {
   ...
}
```

georgeroman (changes requested):

src/sync/events/data/erc721.ts
```diff
@@ -19,6 +21,40 @@ export const transfer: EventData = {
   ]),
 };
 
+export const likeTransfer: EventData = {
+  kind: "erc721-like-transfer",
+  topic: "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
+  addresses: {
+    [CryptoKitties.Addresses.Core[config.chainId]?.toLowerCase()]: true,
+  },
```
vikinatora, line 29: I've added address constraint to the contract we know that emits such event. This is in order to prevent any erc20-like transfer being handled as erc721

src/sync/events/data/erc721.ts
```diff
@@ -19,6 +21,40 @@ export const transfer: EventData = {
...
+    )`,
+  ]),
+};
+
+export const erc20LikeTransfer: EventData = {
+  kind: "erc721-erc20-like-transfer",
+  addresses: {
+    [Beeple.Addresses.Contract[config.chainId]?.toLowerCase()]: true,
+    [CryptoArte.Addresses.Contract[config.chainId]?.toLowerCase()]: true,
+    [CryptoVoxels.Addresses.Parcel[config.chainId]?.toLowerCase()]: true,
```
vikinatora, line 45: Same reason as above

georgeroman:
Let's just remove the old migration files and should be good to go.

vikinatora:
Removed.

PR #2553, opened 2022-12-07
fix: blur issue fix

lljxx1:
### missing sales for bid
- Blur ETH(0x0000000000a39bb272e79075ade125fd351887ac) missing the symbol() 
- added a hardcode when querying the price as WETH

### accurate orderSide
- get call trace from `execute` input data
- and based on the signature to determine orderSide

tests covered 7 cases, need to find the case of multiple bids on the router contract


src/sync/events/handlers/blur.ts
```diff
@@ -41,26 +88,34 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
         }
+
         // Handle: prices
 
         const currency = sell.paymentToken.toLowerCase();
         const currencyPrice = sell.price.div(sell.amount).toString();
+        const isBlurETH = currency ==  "0x0000000000a39bb272e79075ade125fd351887ac";
+
+        // Hardcode as WETH
+        const currencyToPrice = isBlurETH ? Sdk.Common.Addresses.Weth[config.chainId] : currency;
```
georgeroman, line 103: We probably want Eth and not Weth (since Blur bids are basically Eth).

PR #2599, opened 2022-12-11
support element marketplace.

element-market:


src/sync/events/handlers/element.ts
```diff
@@ -73,6 +73,179 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
+        const { args } = eventData.abi.parseLog(log);
+        const orderHash = args["orderHash"].toLowerCase();
+        const maker = args["maker"].toLowerCase();
+        let taker = args["taker"].toLowerCase();
+        const nonce = args["nonce"].toString();
+        const erc20Token = args["erc20Token"].toLowerCase();
+        const erc20TokenAmount = args["erc20TokenAmount"].toString();
+        const erc721Token = args["erc721Token"].toLowerCase();
+        const erc721TokenId = args["erc721TokenId"].toString();
+        const orderId = orderHash + "_" + nonce;
```
georgeroman, line 86: I probably missed this when reviewing the SDK PR but is there any reason to include the nonce in the order id?

element-market: A `BatchSignedOrder` may contains multiple ERC721 items, and each item corresponds to a nonce. The order hash of these items are the same, so they need to be distinguished by orderhash+nonce.

element-market: The `BatchSignedOrder` is stored in the database as multiple orders, and the main difference of these orders is the nonce.

georgeroman: @element-market What do you think of having the order id of such batch-signed orders be the hash of the actual order hash together with the nonce - `keccak256(orderHash, nonce)`? Would that make things more complex (by looking at the code I don't think that should happen)? This would be for keeping the order id format consistent (a 32 byte hash) with the rest of the orders types we support.

element-market: `keccak256(orderHash, nonce)` can also uniquely confirm the order, it's ok to do that.

PR #2602, opened 2022-12-12
feat: order-saving support for nftx

lljxx1:
- create a collection-level buy order for each pool
- for each token in the pool, create a sell order for each token
- using the `mintFee` and `reedemFee` as the fees.


src/utils/nftx/index.ts
```diff
@@ -140,3 +154,113 @@ export const tryParseSwap = async (log: Log) => {
...
+      randomBuy: feeBpsRandomBuy,
+    },
+    currency: WETH,
+    sell: sellPrice,
+    buy: buyPrice,
+    buyRandom: randomBuyPrice,
+  };
+}
+
+export async function getPoolNFTs(vault: string) {
```
georgeroman, line 229: To skip an expensive on-chain call, we should use the off-chain equivalent, as we do here: https://github.com/reservoirprotocol/indexer/blob/5f0527f66728e7620a1db4a2de576186bb0ff8d4/src/orderbook/orders/sudoswap/index.ts#L366

package.json
```diff
@@ -34,7 +34,7 @@
     "@hapi/vision": "^6.1.0",
     "@opensea/stream-js": "^0.0.21-rc.1",
     "@poprank/rankings": "^1.1.20",
-    "@reservoir0x/sdk": "0.0.271",
+    "@reservoir0x/sdk": "^0.0.271",
```
georgeroman, line 37: Let's keep fixed versions, just for safety.

src/utils/nftx/index.ts
```diff
@@ -89,6 +88,15 @@ export const isRedeem = (log: Log, address: string) => {
   }
 };
 
+export const isUserStake = (log: Log) => {
+  if (
+    log.topics[0] === nftx.staked.abi.getEventTopic("UserStaked")
+    // && log.address.toLowerCase() === address
```
georgeroman, line 94: Let's remove this.

src/sync/events/handlers/nftx.ts
```diff
@@ -34,98 +37,108 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
-            } else if (nftPool.address === result.ftPool.token1 && result.amount0Out !== "0") {
-              currency = result.ftPool.token0;
-              currencyPrice = bn(result.amount0Out).div(nftCount).toString();
-            }
-
-            if (currency && currencyPrice) {
-              // Handle: attribution
+        const hasSwap = swapEventsCount > 1;
+
+        if (hasSwap) {
```
georgeroman, line 65: Shouldn't it be the other way around - if we have more than one swap we break?

src/orderbook/orders/nftx/index.ts
```diff
@@ -0,0 +1,510 @@
...
+
+  const handleOrder = async ({ orderParams }: OrderInfo) => {
+    try {
+      const pool = await nftx.getNftPoolDetails(orderParams.pool);
+      if (!pool) {
+        throw new Error("Could not fetch pool details");
+      }
+
+      const amount = 10;
+      const poolPrice = await Sdk.Nftx.helpers.getPoolPrice(
```
georgeroman, line 55: We should assume the buyer always buys a single amount and then get a list of prices for buying each incrementing amount:
`[PriceToBuyQuantity1, PriceToBuyQuantity2, ...]`

georgeroman (changes requested):

package.json
georgeroman: This was not addressed yet.

georgeroman (changes requested):
The below requirements (eg. having a single minted and swap event in the same transaction) were made in order to simplify sale detection logic. It would be very hard to get the price of a sale across multiple transactions. However, now that we're handling NFTX liquidity we cannot afford losing some sale events just because we cannot parse them. What we should do here is make sure that for every sale for which we cannot detect the price, we properly update the fillability status of the impacted orders. The logic would be something like:
```
if (cannotParseSale) {
  updateImpactedOrdersSetFillabilityStatusToFilled;
  addOrdersToOrderUpdatesByIdQueue;
}
```

Here's the sale parsing conditions I initially mentioned:
https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/sync/events/handlers/nftx.ts#L40-L52

src/orderbook/orders/nftx/index.ts
```diff
@@ -0,0 +1,533 @@
...
+  const results: SaveResult[] = [];
+  const orderValues: DbOrder[] = [];
+
+  const handleOrder = async ({ orderParams }: OrderInfo) => {
+    try {
+      const pool = await nftx.getNftPoolDetails(orderParams.pool);
+      if (!pool) {
+        throw new Error("Could not fetch pool details");
+      }
+
```
georgeroman, line 53: For now let's skip ERC20 pools (eg. pools for which the currency is not ETH).

georgeroman:
Let's include the orderId in the fill events. Also, for buy orders we should use fillEventsPartial while for sell orders we should use fillEventsOnChain. Had some issues recently with Sudoswap because of this.

https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/sync/events/handlers/nftx.ts#L99

https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/sync/events/handlers/nftx.ts#L209

See the sudoswap code and how we handle events, it should be exactly the same for NFTX:
https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/sync/events/handlers/sudoswap.ts#L18-L19

georgeroman:
Also, we should be careful about sweeping NFTX listings, since multiple tokens from the same poll would get dynamically priced. We want to have logic similar to Sudoswap:
https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/api/endpoints/execute/get-execute-buy/v6.ts#L208-L221

georgeroman:
Related to the above, we also want to return dynamic pricing information for NFTX listings (eg. the price points), similar to Sudoswap:
https://github.com/reservoirprotocol/indexer/blob/b05cfa714742d6826eb121be159b16c148e2fc94/src/api/endpoints/tokens/get-tokens/v5.ts#L801-L820

src/orderbook/orders/nftx/index.ts
lljxx1: https://github.com/lljxx1/nft-exchanges/blob/main/nftx-protocol-v2/contracts/solidity/NFTXStakingZap.sol#L424

lljxx1: all pools are WETH

PR #2609, opened 2022-12-12
Add Tofu NFT sales ingestion

vikinatora:
SDK PR: https://github.com/reservoirprotocol/core/pull/65
I've left attribution data logic inside the handler logic. That's for future compatability in case we implement router execution for Tofu NFT

Let me know if `tofu` isn't descriptive enough and I can swap it out for a different name.


src/sync/events/data/tofu.ts
```diff
@@ -0,0 +1,27 @@
+import { Interface } from "@ethersproject/abi";
+
+import { config } from "@/config/index";
+import { EventData } from "@/events-sync/data";
+import { TofuNft } from "@reservoir0x/sdk";
+
+export const inventoryUpdate: EventData = {
+  kind: "tofu-inventory-update",
+  addresses: { [TofuNft.Addresses.Exchange[config.chainId]?.toLowerCase()]: true },
```
georgeroman, line 9: We'll need an SDK integration as well.

georgeroman (changes requested):

src/sync/events/data/tofu.ts
vikinatora: Yes, forgot to publish the PR: https://github.com/reservoirprotocol/core/pull/65

georgeroman: Cool, merged and published in version `0.0.268`. Let's update `package.json` and `yarn.lock` with the new version for this PR. Also, on a second thought, let's call the order kind `tofu-nft` instead of `tofu`.

vikinatora: Done

vikinatora: Done

PR #2622, opened 2022-12-13
feat: add currency filter to tokens endpoint

jasonyjong:
Support `currencies` query arg for `/tokens/v5` endpoint.

We refer to currencies by their contract address. (e.g. ETH has address `0x0000000000000000000000000000000000000000`)

Optionally pass in either a single string address or an array of addresses.

When the currency filter is passed in we return the floor price of listings in the filtered currency, similar to what we do with the `source` filter.

Open questions:
1. Is address the right way to refer to a currency?  Should we consider filtering on symbol, like ETH, APE?
2. Is the right query arg `currency` or `currencies`?


jasonyjong:
@ipeleg - let me know the best way to get reviews on your backlog!

jasonyjong:
@ipeleg - not sure how important indexes are here, but I added some that should be relevant if we choose that route.

ipeleg (approved):

PR #2666, opened 2022-12-15
feat: add price filter to tokens endpoint

jasonyjong:
Add minFloorAskPrice & maxFloorAskPrice to `/tokens/v5` endpoint

Unit is floor sell price converted to ETH.

Open question:
1. Is decimal ETH the right unit? Should we use wei instead?  Passing in ETH does lead to confusing rounding phenomena (e.g. a listing with price 0.11111, with price rounded to 0.11, will be filtered out if filter of max=0.11 is passed in)
2. Is `floorAskPrice` the right terminology to use.


jasonyjong:
@ipeleg - let me know how this looks!  I believe the sort indexes in floor price should be sufficient

PR #2678, opened 2022-12-15
Nft trader - sales events

boris-bonin:
This implementation is only tracking sales events for NftTrader.io integration.
With this PR we only keep track of ERC20/ETH <-> ERC721/ERC1155 trades.

SDK PR: https://github.com/reservoirprotocol/core/pull/67


georgeroman:
Let's resolve the conflicts and then should be good to merge.

PR #2701, opened 2022-12-16
fix: rarity rank filter check undefined

jasonyjong:
Passing in min/maxRarityRank = 0 leads to a bug

@ipeleg


jasonyjong:
@ipeleg good call, updated

PR #2708, opened 2022-12-16
Add Okex sales handling

vikinatora:
Okex exchange contract isn't verified on etherscan but I noticed that it's event hash and event topics count matche the ones from seaport and noun so I decided to reuse the event handling logic from them. Tested with a few sale transactions and it seems to me that the event are being parsed correctly.

SDK PR: https://github.com/reservoirprotocol/core/pull/68

### DIFF
--- a/src/sync/events/index.ts
+++ b/src/sync/events/index.ts
@@ -296,6 +296,10 @@ export const syncEvents = async (
       {
         kind: "nft-trader",
         events: enhancedEvents.filter(({ kind }) => kind.startsWith("nft-trader")),
+      },
+      {
+        kind: "okex",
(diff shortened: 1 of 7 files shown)


georgeroman:
Cool, probably just a Seaport fork

src/sync/events/index.ts
```diff
@@ -296,6 +296,10 @@ export const syncEvents = async (
       {
         kind: "nft-trader",
         events: enhancedEvents.filter(({ kind }) => kind.startsWith("nft-trader")),
+      },
+      {
+        kind: "okex",
+        events: enhancedEvents.filter(({ kind }) => kind.startsWith("okex")),
```
vikinatora, line 302: Will fix this in a sec

PR #2781, opened 2022-12-22
Zeroex v2

boris-bonin:
0x v2 sales integration. 
Similar to NFT Trader, in this integration we avoid sales that are other than 1 erc721/1155 <-> eth/erc20 and bundles.

sdk pr: https://github.com/reservoirprotocol/core/pull/73

### DIFF
--- a/src/orderbook/orders/index.ts
+++ b/src/orderbook/orders/index.ts
@@ -164,8 +164,6 @@ export const getOrderSourceByOrderKind = async (
         return sources.getOrInsert("benddao.xyz");
       case "superrare":
         return sources.getOrInsert("superrare.com");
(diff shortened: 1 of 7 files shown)


src/orderbook/orders/index.ts
```diff
@@ -163,6 +164,8 @@ export const getOrderSourceByOrderKind = async (
         return sources.getOrInsert("benddao.xyz");
       case "superrare":
         return sources.getOrInsert("superrare.com");
+      case "zeroex-v2":
```
georgeroman, line 167: Let's remove this, the source should be unknown for zeroex v2 since multiple parties used them under the hood.

PR #2815, opened 2022-12-27
feat: add seaport royalties

lljxx1:


src/utils/orders.ts
```diff
@@ -70,12 +70,7 @@ export class Orders {
                       json_build_object(
                         'kind', 'custom'
                       )
-                    )
-                  WHEN token_sets.attribute_id IS NULL THEN
```
georgeroman, line 74: I don't think this change belongs here

PR #2839, opened 2022-12-29
feat: pack source to the seaport salt

lljxx1:


src/orderbook/orders/seaport/build/utils.ts
```diff
@@ -23,13 +25,22 @@ export interface BaseOrderBuildOptions {
...
 
 type OrderBuildInfo = {
   params: BaseBuildParams;
   kind: "erc721" | "erc1155";
 };
 
+export function padSourceToSalt(source: string, salt?: string) {
+  const sourceHash = generateSourceBytes(source);
+  const saltHex = BigNumber.from(salt)._hex.slice(2);
+  const padSourceHex = `0x${sourceHash}${saltHex}`;
```
georgeroman, line 39: If `saltHex` is greater than 28 bytes, this will fail. What we should do instead is simply override the first 4 bytes of the salt with the domain hash, rather than concatenating.

georgeroman (changes requested):

PR #2845, opened 2022-12-29
feat: filter by collection set in user top bids api

JadenDurnford:


src/api/endpoints/orders/get-user-top-bids/v3.ts
```diff
@@ -175,6 +180,15 @@ export const getUserTopBidsV3Options: RouteOptions = {
       communityFilter = `AND community = $/community/`;
     }
 
+    if (query.collectionsSetId) {
+      query.collectionsIds = await CollectionSets.getCollectionsIds(query.collectionsSetId);
+      if (_.isEmpty(query.collectionsIds)) {
+        throw Boom.badRequest(`No collections for collection set ${query.collectionsSetId}`);
+      }
+
+      collectionSetFilter = `AND id IN ($/collectionsIds:csv/)`;
```
nofir, line 189: you can simplify it and just set the collectionFilter because you are just getting a list of collection ids (which is similar to passing multiple collection ids in the request.

nofir (changes requested):

PR #2854, opened 2022-12-30
Jaden/res 1586 opensea attribute bidding

JadenDurnford:


src/jobs/orderbook/post-order-external/index.ts
```diff
@@ -213,29 +213,65 @@ const postOrder = async (
         order.getInfo()?.side === "buy" &&
         ["contract-wide", "token-list"].includes(order.params.kind!)
       ) {
-        const { collectionSlug } = await redb.oneOrNone(
+        const { collectionInfo } = await redb.oneOrNone(
```
nofir, line 216: I think you need to have an alias on the return to be able extract as `collectionInfo`. Also, `orderKindInfo` might be a better name now

src/jobs/orderbook/post-order-external/index.ts
```diff
@@ -213,29 +213,65 @@ const postOrder = async (
...
+        if (!collectionInfo) {
+          throw new Error("Invalid collection/trait offer.");
         }
 
-        return OpenSeaApi.postCollectionOffer(order, collectionSlug, orderbookApiKey);
+        return OpenSeaApi.postCollectionOrTraitOffer(
+          order,
+          collectionInfo.collection_slug,
+          orderbookApiKey,
+          collectionInfo?.attribute ?? null
```
nofir, line 273: I don't think you need `?? null`

src/orderbook/orders/seaport/build/buy/attribute.ts
```diff
@@ -61,11 +62,28 @@ export const build = async (options: BuildOrderOptions) => {
...
+        attributeResult.slug,
+        options.attributes[0].key,
+        options.attributes[0].value
+      );
 
-    // Fetch all tokens matching the attributes
-    const tokens = await redb.manyOrNone(
-      `
+      if (
+        buildCollectionOfferParams.partialParameters.consideration[0].identifierOrCriteria != "0"
```
nofir, line 75: probably doesnt matter but i dont think this is ever an option in the case of trait offers.

src/orderbook/orders/seaport/index.ts
```diff
@@ -370,14 +371,73 @@ export const save = async (
             // This is due to a mismatch between the collection flags in our system and OpenSea.
```
nofir, line 371: this comment should move into the else

src/orderbook/orders/seaport/index.ts
```diff
@@ -370,14 +371,73 @@ export const save = async (
...
-              tokenSetId = `contract:${info.contract}`;
-              await tokenSet.contractWide.save([
-                {
-                  id: tokenSetId,
-                  schemaHash,
-                  contract: info.contract,
-                },
-              ]);
+              if (metadata.schema?.kind === "attribute") {
+                const attribute = await Attributes.getAttributeByCollectionKeyValue(
```
nofir, line 375: Why do you need to get the attribute?

```
SELECT
              token_attributes.token_id
            FROM token_attributes
            JOIN attributes
              ON token_attributes.attribute_id = attributes.id
            JOIN attribute_keys
              ON attributes.attribute_key_id = attribute_keys.id
            JOIN tokens
              ON token_attributes.contract = tokens.contract
              AND token_attributes.token_id = tokens.token_id
            WHERE attribute_keys.collection_id = $/collection/
              AND attribute_keys.key = $/key/
              AND attributes.value = $/value/
```

src/orderbook/orders/seaport/index.ts
```diff
@@ -370,14 +371,73 @@ export const save = async (
...
+
+                const limit = 5000;
+                let checkForMore = true;
+                let continuation = "";
+
+                let tokens: { tokenId: string; isFlagged: number }[] = [];
+
+                while (checkForMore) {
+                  const query = `
+                    SELECT token_attributes.token_id, tokens.is_flagged
```
nofir, line 389: the decision if to take only non flagged tokens should be based if the schema has `isNonFlagged` (see post order v3)

src/orderbook/orders/seaport/index.ts
```diff
@@ -370,14 +371,73 @@ export const save = async (
...
-                },
-              ]);
+              if (metadata.schema?.kind === "attribute") {
+                const attribute = await Attributes.getAttributeByCollectionKeyValue(
+                  metadata.schema.data.collection,
+                  metadata.schema.data.attributes[0].key,
+                  metadata.schema.data.attributes[0].value
+                );
+
+                const limit = 5000;
```
nofir, line 381: I would move the entire pagination logic to get the attribute tokens to a separate function

src/orderbook/orders/seaport/index.ts
nofir: otherwise, the` tokenList.save` could fail because its validating based on the schema

nofir (changes requested):

nofir:
will create a new pr

PR #2855, opened 2023-01-02
feat: royalties support more marketplace

lljxx1:
- make the `extractRoyalties` can work with every kind of`fillEvent`
- a config `platformFeeRecipientsRegistry`added to store all marketplace-related fee recipient's address
- `paidFullRoyalty` change to check the bps diff

#### tests
- `x2y2` and `looks-rare` multiple scenes tested


georgeroman:
Let's remove the two committed `.DS_Store` files and then should be good to go.

PR #2860, opened 2023-01-03
(feat) RES-1527 - Add bids total amount

mplelis:


georgeroman:
@ipeleg Asking for your review on this one, does the `SUM() OVER()` make sense here?

PR #2883, opened 2023-01-05
feat: add nftx

lljxx1:


src/sync/events/handlers/nftx.ts
```diff
@@ -5,14 +5,22 @@ import * as es from "@/events-sync/storage";
 import * as utils from "@/events-sync/utils";
 import { getUSDAndNativePrices } from "@/utils/prices";
 import * as nftxUtils from "@/utils/nftx";
+import { redb } from "@/common/db";
```
georgeroman, line 8: Let's use `idb` instead of `redb` so that there won't be any data propagation delay issues.

src/sync/events/handlers/nftx.ts
```diff
@@ -236,19 +362,147 @@ export const handleEvents = async (events: EnhancedEvent[]): Promise<OnChainData
...
+          nftx.getOrderId(baseEventParams.address, "sell", tokenId)
+        );
+        const dbOrders = await redb.manyOrNone(
+          `SELECT * FROM orders where id IN ($/orderIds:list/)`,
+          {
+            orderIds,
+          }
+        );
+
+        for (let index = 0; index < tokenIds.length; index++) {
```
georgeroman, line 426: The same comment above applies here, all we want to do is mark the orders as filled, no fill event insertion needed.

georgeroman (changes requested):

georgeroman:
Also, let's fix the conflicts and update the branch to merge into from v5-dev to development. We changed v5-dev -> development and v5 -> main a few days ago.

src/sync/events/storage/fill-events/index.ts
```diff
@@ -26,6 +26,7 @@ export type Event = {
   isPrimary?: boolean;
   baseEventParams: BaseEventParams;
 
+  updateOrderOnly?: boolean;
```
georgeroman, line 29: Let's remove this, I assume it's just a relic.

PR #2915, opened 2023-01-09
feat: add royalty details to the sales api

lljxx1:


src/common/joi.ts
```diff
@@ -3,6 +3,7 @@ import Joi from "joi";
 import { formatEth, formatPrice, formatUsd, now, regex } from "@/common/utils";
 import { Currency, getCurrency } from "@/utils/currencies";
 import { getUSDAndNativePrices } from "@/utils/prices";
+import { bn } from "@reservoir0x/sdk/dist/utils";
```
georgeroman, line 6: You can get `bn` from `@/common/utils`.

src/common/joi.ts
```diff
@@ -26,11 +27,16 @@ export const JoiPrice = Joi.object({
   netAmount: JoiPriceAmount.optional(),
 });
 
+export const subFeeWithBps = (amount: string | number, totalFeeBps: number) => {
```
georgeroman, line 30: Do we need to export this? We need it somewhere else?

georgeroman, line 30: `amount` can maybe have a `BigNumberish` type?

src/common/joi.ts
```diff
@@ -84,7 +99,8 @@ export const getJoiPriceObject = async (
         currency,
         prices.net.amount,
         prices.net.nativeAmount,
-        prices.net.usdAmount
```
georgeroman, line 87: Let's have the following logic in this method:
- if `prices.net` is passed use that
- if `prices.net` is not passed but `totalFeeBps` is passed, compute the net amount from the gross amount by subtracting the fee
- otherwise, do nothing (don't return net amount basically)

georgeroman (changes requested):

PR #2947, opened 2023-01-10
Jaden res 1373 no merge

JadenDurnford:


nofir:
Implemented as part of https://github.com/reservoirprotocol/indexer/pull/2985

PR #3142, opened 2023-01-24
fix: remove indexes unused by both writer and reader

tv3636:


src/migrations/1674595177_unused-indexes.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+
+DROP index orders_side_contract_created_at_index, tokens_collection_id_source_id_floor_sell_value_index, token_normalized_floor_sell_events_contract_token_id_created_at, collections_normalized_floor_sell_value_index, collections_non_flagged_floor_sell_value_index, user_activities_bid_no_collection_index, activities_bid_no_collection_index, orders_offer_bundle_id_index, orders_consideration_bundle_id_index, bundle_items_token_set_id_bundle_id_index;
```
georgeroman, line 3: Can we keep `orders_side_contract_created_at_index`, just added this one and it's for a specific use-case (which no one is using at the moment but might do in the near future)?

ipeleg (approved):

ipeleg:
@tv3636 Looks good!

PR #3168, opened 2023-01-27
feat: update token flag status on opensea synchronous checks

georgeroman:


nofir (approved):
Looks good! 

Maybe we need to think about how to consolidate the the logic [here](https://github.com/reservoirprotocol/indexer/blob/241b2cca011ba71230f256b5279ad65aa494f389/src/jobs/metadata-index/write-queue.ts#L126) after detecting a flag status change.

PR #3216, opened 2023-01-31
fix: update trace to include POST request payload

tv3636:


src/models/api-keys/index.ts
```diff
@@ -210,6 +211,16 @@ export class ApiKeyManager {
   public static async logRequest(request: Request) {
     const log: any = await ApiKeyManager.getBaseLog(request);
     logger.info("metrics", JSON.stringify(log));
+
+    // Add request query or payload to Datadog trace
+    try {
+      tracer
+        .scope()
+        .active()
+        ?.setTag("query", log.payload ? log.payload : log.query ? log.query : null);
```
nofir, line 220: you can potentially have all 3, query, payload, params. you can set one tag - something like request, that has all 3:

    if (request.payload) {
      log.payload = request.payload;
    }
    if (request.params) {
      log.params = request.params;
    }
    if (request.query) {
      log.query = request.query;
    }

tv3636: merged params into the same object - assuming there are not conflicts between params and query/payload

nofir: I think it should be simpler: (Request is already used in the trace today so i named it "requestParams", but maybe you have a better name...)
 

tracer
        .scope()
        .active()
        ?.setTag(
          "requestParams",
          {
             payload: request.payload,
             params: request. params,
             query: request.query
          }
        );

tv3636: I'm thinking we'd want them merged because it simplifies filtering and grouping in Datadog. 

Filtering will be more convenient since we can filter on a single key like `requestParams.collection` rather than three separate ones each time.

If we wanted to group by `collection` to see for instance average latency for all requests by collection value, it becomes pretty complicated in Datadog to do this if we have `requestParams.payload.collection`, `requestParams.params.collection`, and `requestParams.query.collection` (but maybe this is an impractical use case? I'm not sure which keys this may be useful for if any)

We can also tell where the values came from based on the url and request method, so I'd advocate combining them at least for convenience

nofir: yeah, i see what you mean. i dont mind playing around with it on goerli. my point is to have it consistent with other places with log the request.

PR #3266, opened 2023-02-07
feat: support buy with erc20

lljxx1:


src/utils/permits/erc20.ts
```diff
@@ -0,0 +1,16 @@
+import { keccak256 } from "@ethersproject/solidity";
+import { FTPermit } from "@reservoir0x/sdk/dist/router/v6/types";
+import stringify from "json-stable-stringify";
+
+import { redis } from "@/common/redis";
+
+export const getPermitId = (requestPayload: object, tokens: string[]) =>
+  `erc20-permit:${keccak256(["string"], [stringify({ requestPayload, tokens })])}`;
```
georgeroman, line 8: Let's rename this from `erc20-permit` to `ft-permit` to be consistency with the rest of the codebase. Also let's rename the file to `ft.ts`.

src/api/endpoints/execute/get-execute-buy/v6.ts
```diff
@@ -708,6 +719,87 @@ export const getExecuteBuyV6Options: RouteOptions = {
...
+        : undefined;
+
+      const inputCurrencyAmount = payload.amount ? bn(payload.amount) : undefined;
+
+      for (const approval of approvals) {
+        const approvedValue = await onChainData
+          .fetchAndUpdateFtApproval(
+            approval.currency,
+            approval.operator,
+            Sdk.Common.Addresses.Permit2[config.chainId],
```
georgeroman, line 737: This seems wrong. We should use all fields from the `approval` object, like we do here: https://github.com/reservoirprotocol/indexer/blob/1c2569f4c0c679e1198d2c1d206a4866d96b91f4/src/api/endpoints/execute/get-execute-sell/v7.ts#L606-L608

src/api/endpoints/execute/get-execute-buy/v6.ts
```diff
@@ -708,6 +719,87 @@ export const getExecuteBuyV6Options: RouteOptions = {
...
+            true
+          )
+          .then((a) => a.value);
+
+        // Make sure been approved to Permit2
+        const isApproved = inputCurrencyAmount
```
georgeroman, line 743: The approval check is wrong. What we should do instead is the the total amount to get approved for the current currency by simply summing the permit amounts of that particular currency. A pseudo-code for this might look like:
```
   const amountToApprove = permits.find(p => p.currency === approval.currency).sum(p => p.amount);
   const isApproved = bn(approvedValue).gte(amountToApprove);
```

src/api/endpoints/execute/get-execute-buy/v6.ts
```diff
@@ -5,13 +5,17 @@ import * as Boom from "@hapi/boom";
 import { Request, RouteOptions } from "@hapi/hapi";
```
georgeroman, line 5: This is a breaking change so we should leave the currency `v6` API as it is and move all the changes to a new `v7` API.

georgeroman (changes requested):

src/api/endpoints/execute/get-execute-buy/v7.ts
```diff
@@ -0,0 +1,878 @@
...
+        }
+      } else {
+        const erc20 = new Sdk.Common.Helpers.Erc20(baseProvider, buyInCurrency);
+
+        const balance = await erc20.getBalance(txSender);
+        if (!payload.skipBalanceCheck && bn(balance).lt(totalPrice)) {
+          throw Boom.badData("Balance too low to proceed with transaction");
+        }
+
+        let conduit: string;
```
georgeroman, line 818: We should remove this approval check and generation since we already handled it above.

src/api/endpoints/execute/get-execute-buy/v7.ts
```diff
@@ -0,0 +1,878 @@
...
+import * as commonHelpers from "@/orderbook/orders/common/helpers";
+import * as sudoswap from "@/orderbook/orders/sudoswap";
+import * as nftx from "@/orderbook/orders/nftx";
+import { getCurrency } from "@/utils/currencies";
+
+const version = "v7";
+
+export const getExecuteBuyV7Options: RouteOptions = {
+  description: "Buy tokens",
+  tags: ["api", "Router"],
```
georgeroman, line 31: For now let's use the `["api", "x-experimental"]` tag.

georgeroman (changes requested):

PR #3273, opened 2023-02-07
feat: adding api details admin endpoint

JadenDurnford:


src/api/routes.ts
```diff
@@ -307,6 +307,12 @@ export const setupRoutes = (server: Server) => {
     options: apiKeysEndpoints.getApuKeyRateLimits,
   });
 
+  server.route({
+    method: "GET",
+    path: "/api-keys/{key}/details",
```
ipeleg, line 312: Let's add admin prefix to the route

src/api/endpoints/api-keys/get-api-key-details.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ipeleg, line 1: Move this file to the admin endpoints folder

src/api/routes.ts
```diff
@@ -293,6 +293,12 @@ export const setupRoutes = (server: Server) => {
     options: adminEndpoints.postRoutersOptions,
   });
 
+  server.route({
+    method: "GET",
+    path: "/admin/{key}/details",
```
nofir, line 298: I think this should be `/admin/get-api-key` which follows the convention of other routes.

src/api/endpoints/admin/get-api-key-details.ts
```diff
@@ -0,0 +1,62 @@
...
+  response: {
+    schema: Joi.object({
+      key: Joi.string().uuid(),
+      appName: Joi.string(),
+      website: Joi.string(),
+      email: Joi.string().email(),
+      active: Joi.bool(),
+      tier: Joi.number().unsafe(),
+      permissions: Joi.string().allow(null),
+      createdAt: Joi.string().allow(null),
```
nofir, line 30: can this actually be null?

src/api/endpoints/admin/get-api-key-details.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      return {
+        key: apiKey?.key,
+        appName: apiKey?.appName,
+        website: apiKey?.website,
+        email: apiKey?.email,
+        active: apiKey?.active,
+        tier: apiKey?.tier,
+        permissions: apiKey?.permissions,
+        createdAt: apiKey?.createdAt.toString(),
```
nofir, line 55: new Date(apiKey.createdAt).toISOString()

src/api/endpoints/admin/get-api-key-details.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw Boom.unauthorized("Wrong or missing admin API key");
+    }
+
+    const params = request.params as any;
+
+    try {
+      const apiKey = await ApiKeyManager.getApiKey(params.key);
+
+      return {
+        key: apiKey?.key,
```
nofir, line 48: its cleaner to return an error when the key isn't found rather than returning an empty object with 200 http status and using `apiKey?`

nofir (changes requested):

src/api/endpoints/admin/get-api-key-details.ts
JadenDurnford: In theory it always defaults to the current time, but the column is still nullable so I added that just in case

nofir: Any reason the column is nullable?

nofir (approved):

PR #3276, opened 2023-02-08
fix: use decoded zeroex-v2 asset data instead of trace

boris-bonin:
SDK: https://github.com/reservoirprotocol/core/pull/82

### DIFF
--- a/src/sync/events/handlers/zeroex-v2.ts
+++ b/src/sync/events/handlers/zeroex-v2.ts
@@ -84,7 +84,7 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
         tokenContract = orderSide === "sell" ? decodedMakerAssetData[0] : decodedTakerAssetData[0];
         tokenId = orderSide === "sell" ? decodedMakerAssetData[1] : decodedTakerAssetData[1];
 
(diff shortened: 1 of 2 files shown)


src/sync/events/handlers/zeroex-v2.ts
```diff
@@ -15,42 +15,79 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+          );
+        }
+
+        currency = orderSide === "sell" ? decodedTakerAssetData[0] : decodedMakerAssetData[0];
+        currencyPrice = orderSide === "sell" ? takerAssetData[0][0] : makerAssetData[0][0];
+        amount = orderSide === "sell" ? makerAssetData[0][0] : takerAssetData[0][0];
+        tokenContract = orderSide === "sell" ? decodedMakerAssetData[0] : decodedTakerAssetData[0];
+        tokenId = orderSide === "sell" ? decodedMakerAssetData[1] : decodedTakerAssetData[1];
+
+        if (currency === Sdk.ZeroExV2.Addresses.Eth[config.chainId]) {
```
georgeroman, line 87: Let's use the Eth address from ZeroExV4

PR #3315, opened 2023-02-13
feat: handle custom token with permit when execute bid

lljxx1:


package.json
```diff
@@ -34,7 +34,7 @@
     "@hapi/vision": "^6.1.0",
     "@opensea/stream-js": "^0.0.21-rc.1",
     "@poprank/rankings": "^1.1.20",
-    "@reservoir0x/sdk": "0.0.312",
+    "@reservoir0x/sdk": "/Users/fun/projects/reservoir/core/packages/sdk",
```
georgeroman, line 37: Leftover from local testing?

yarn.lock
```diff
@@ -2675,6 +2675,14 @@
   resolved "https://registry.npmjs.org/@protobufjs/utf8/-/utf8-1.1.0.tgz"
   integrity sha512-Vvn3zZrhQZkkBE8LSuW3em98c0FwgO4nxzv6OdSxPKJIEKY2bGbHn+mhGIPerzI4twdxaP8/0+06HBpwf345Lw==
 
+"@reservoir0x/sdk@/Users/fun/projects/reservoir/core/packages/sdk":
```
georgeroman, line 2678: Leftover from local testing?

src/api/endpoints/execute/get-execute-sell/v7.ts
```diff
@@ -540,16 +541,21 @@ export const getExecuteSellV7Options: RouteOptions = {
...
       }
 
+      const { customTokenAddresses } = getNetworkSettings();
       const router = new Sdk.RouterV6.Router(config.chainId, baseProvider, {
         x2y2ApiKey: payload.x2y2ApiKey ?? config.x2y2ApiKey,
         cbApiKey: config.cbApiKey,
       });
+
+      const forcePermit =
+        bidDetails.length === 1 && customTokenAddresses.includes(bidDetails[0].contract);
```
georgeroman, line 551: Do we need the `bidDetails.length === 1` check here? Seems redundant since the force permit option won't even be checked if we have multiple bids to fill.

georgeroman (changes requested):

PR #3333, opened 2023-02-14
Manifold accept offers

boris-bonin:


src/sync/events/handlers/manifold.ts
```diff
@@ -367,6 +367,131 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+          for (const token of tokenPayouts) {
+            const tokensTransfered = bn(state[payoutAddress].tokenBalanceState[token]);
+            if (tokensTransfered.gt(0) && (!maxPayout || tokensTransfered.gt(maxPayout))) {
+              maxPayout = tokensTransfered;
+              taker = payoutAddress;
+            }
+          }
+        }
+
+        if (!taker) {
```
georgeroman, line 433: Let's check every case: `tokenId`, `amount`, `tokenContract`, like we do here: https://github.com/reservoirprotocol/indexer/blob/4f8b071dc7ef767d287c8f88e71d051b5b0ef5ea/src/sync/events/handlers/manifold.ts#L304-L315

PR #3355, opened 2023-02-16
feat: update flightcontrol.json to configure projects from code

tv3636:
Note this change does not impact existing projects in flightcontrol, but will enable us to create new duplicate projects which are built from `flightcontrol.json` (with the goal of re-pointing cloudflare to these duplicates to fully switch over)

Using `flightcontrol.json` enables us to track revision history of project configuration and also simplifies/enhances our Datadog agent setup, which is currently limited on railway. Incidentally this also fixes profiling in Datadog

### DIFF
--- a/flightcontrol.json
+++ b/flightcontrol.json
@@ -1,151 +1,244 @@
 {
   "environments": [
     {
-      "id": "production",
-      "name": "Production",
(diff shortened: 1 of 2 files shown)


nofir:
@tv3636 can this be closed?

PR #3370, opened 2023-02-17
fix: add contract filtering on activity apis

mplelis:


packages/indexer/src/api/endpoints/activities/get-activities/v1.ts
```diff
@@ -59,7 +65,11 @@ export const getActivityV1Options: RouteOptions = {
     const query = request.query as any;
 
     try {
-      const activities = await Activities.getActivities(query.continuation, query.limit);
+      const activities = await Activities.getActivities({
+        continuation: query.continuation,
+        limit: query.limit,
+        contract: query.contract ?? undefined,
```
nofir, line 71: Not sure "?? undefined" is necessary, as it will be undefined if not passed in the quest

packages/indexer/src/models/activities/index.ts
```diff
@@ -203,19 +212,32 @@ export class Activities {
...
       }
 
       baseQuery += ` ORDER BY event_timestamp ${sortDirection}, id ${sortDirection}`;
     } else {
       if (!_.isNull(continuation) && continuation !== "null") {
         id = continuation;
         const sign = sortDirection == "desc" ? "<" : ">";
-        baseQuery += ` WHERE id ${sign} $/id/`;
+        conditions.push(`id ${sign} $/id/`);
+        baseQuery += " WHERE " + conditions.map((c) => `(${c})`).join(" AND ");
```
nofir, line 237: wouldn't that break the query?  you are already doing that on line 240

nofir (changes requested):

nofir (dismissed):

nofir:
will create a new pr

PR #3404, opened 2023-02-23
feat: support seaport on-chain liquidlity

lljxx1:


packages/indexer/src/orderbook/orders/seaport/index.ts
```diff
@@ -202,14 +202,16 @@ export const save = async (
...
 
-      // Check: order has a valid signature
-      try {
-        await order.checkSignature(baseProvider);
-      } catch {
-        return results.push({
-          id,
-          status: "invalid-signature",
-        });
+      // Check: order has a valid signature aand skip on-chain orders
```
georgeroman, line 205: Let's change this comment to `Check: order has a valid signature (or was validated on-chain)`

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -209,6 +210,50 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+        const orderId = parsedLog.args["orderHash"].toLowerCase();
+
+        const tx = await utils.fetchTransaction(baseEventParams.txHash);
+        const orderKind = subKind.startsWith("seaport-v1.2") ? "seaport-v1.2" : "seaport";
+
+        const isV12 = orderKind === "seaport-v1.2";
+        const exchange = isV12
+          ? new Sdk.SeaportV12.Exchange(config.chainId)
+          : new Sdk.Seaport.Exchange(config.chainId);
+        const inoutData = exchange.contract.interface.decodeFunctionData("validate", tx.data);
```
georgeroman, line 226: Typo: should be `inputData`.

georgeroman: Also, we should have different logic depending on the event kind (v1.1 or v1.4). If the version is v1.1 then we should get the order from the calldata. However, we should support internal calls as well so we should do it by parsing the call trace. If the version is v1.4 then we already have the order in the event and we should simply use that.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -209,6 +210,50 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+        const inputOrders = inoutData.orders;
+        for (let index = 0; index < inputOrders.length; index++) {
+          const orderData = inputOrders[index];
+          try {
+            const { parameters } = orderData;
+            const counter = await exchange.getCounter(baseProvider, parameters.offerer);
+            const order = new Sdk.Seaport.Order(config.chainId, {
+              kind: "single-token",
```
georgeroman, line 234: We shouldn't assume an order is `single-token`. Let's just remove the `kind` and make sure to wrap the order creation in a `try / catch` and ignore any errors.

georgeroman (changes requested):

packages/indexer/src/sync/events/data/seaport-v1.4.ts
```diff
@@ -61,3 +61,42 @@ export const counterIncremented: EventData = {
     )`,
   ]),
 };
+
+export const orderValidated: EventData = {
+  kind: "seaport",
+  subKind: "seaport-v1.2-order-validated",
```
georgeroman, line 67: The sub-kind should be `seaport-v1.4-order-validated`.

packages/indexer/src/sync/events/data/seaport-v1.4.ts
```diff
@@ -61,3 +61,42 @@ export const counterIncremented: EventData = {
...
+  addresses: { [SeaportV12.Addresses.Exchange[config.chainId]?.toLowerCase()]: true },
```
georgeroman, line 68: Does this compile? Should be `SeaportV14`.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -209,6 +211,85 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+              allOrderParameters.push(inputData.orders[index].parameters);
+            }
+          }
+        }
+
+        for (let index = 0; index < allOrderParameters.length; index++) {
+          const parameters = allOrderParameters[index];
+          try {
+            const counter = await exchange.getCounter(baseProvider, parameters.offerer);
+            const order = new Sdk.Seaport.Order(config.chainId, {
```
georgeroman, line 267: We should differentiate between `seaport` and `seaport-v1.4` parameters. Maybe have separate arrays for each?

packages/indexer/src/orderbook/orders/seaport-v1.4/index.ts
```diff
@@ -203,13 +203,15 @@ export const save = async (
...
 
       // Check: order has a valid signature
-      try {
-        await order.checkSignature(baseProvider);
-      } catch {
-        return results.push({
-          id,
-          status: "invalid-signature",
-        });
+      if (!order.params.onChain) {
```
georgeroman, line 206: If we do it this way (by passing an `onChain` field in the order params) anyone could post orders with that field enabled and we will skip signature validation. What we should do instead is pass the `onChain` field in the `OrderInfo` object so that we as the backend have full control over when that's set or not. We should remove the `onChain` field from the SDK types.

georgeroman (changes requested):

packages/indexer/src/orderbook/orders/seaport-v1.4/index.ts
lljxx1: move to the metadata

packages/indexer/src/sync/events/data/index.ts
```diff
@@ -218,6 +220,7 @@ const allEventData = [
   seaport.counterIncremented,
   seaport.orderCancelled,
   seaport.orderFulfilled,
+  seaport.orderValidated,
   seaportV14.counterIncremented,
   seaportV14.orderCancelled,
   seaportV14.orderFulfilled,
```
georgeroman, line 226: We also need to add `seaportV14.orderValidated` here.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -240,6 +242,121 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+            const counter = await exchange.getCounter(baseProvider, parameters.offerer);
+            const order = new Sdk.SeaportV14.Order(config.chainId, {
+              ...parameters,
+              onChain: true,
+              counter,
+            });
+            order.params.signature = HashZero;
+            // Order hash match
+            if (orderId === order.hash()) {
+              // onChainData.orders.push({
```
georgeroman, line 340: We should uncomment this.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -240,6 +242,121 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+            );
+            if (matchCall) {
+              validateCalls.push(matchCall);
+            } else {
+              break;
+            }
+          }
+
+          for (let index = 0; index < validateCalls.length; index++) {
+            const inputData = exchange.contract.interface.decodeFunctionData(
```
georgeroman, line 286: Should we wrap this in a `try`/`catch`? Otherwise any failed call will stop the whole flow, which isn't what we want.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -240,6 +242,125 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+              });
+              // Skip
+              break;
+            }
+          } catch (error) {
+            // parse error
+          }
+        }
+
+        // Hanlde seaport-v1.4
```
georgeroman, line 331: Typo: `Hanlde` instead of `Handle`. Same above for handling Seaport orders.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -218,6 +242,125 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+        }
+
+        // Handle seaport
+        for (let index = 0; index < allOrderParameters.length; index++) {
+          const parameters = allOrderParameters[index];
+          try {
+            const counter = await exchange.getCounter(baseProvider, parameters.offerer);
+            const order = new Sdk.Seaport.Order(config.chainId, {
+              ...parameters,
+              onChain: true,
```
georgeroman, line 307: Let's remove this.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -218,6 +242,125 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+        }
+
+        // Handle seaport-v1.4
+        for (let index = 0; index < allOrderParametersV14.length; index++) {
+          const parameters = allOrderParametersV14[index];
+          try {
+            const counter = await exchange.getCounter(baseProvider, parameters.offerer);
+            const order = new Sdk.SeaportV14.Order(config.chainId, {
+              ...parameters,
+              onChain: true,
```
georgeroman, line 338: Let's remove this.

PR #3426, opened 2023-02-24
feat: prepare for seaport-v1.4

georgeroman:


nofir (dismissed):

packages/indexer/src/websockets/opensea/index.ts
```diff
@@ -56,12 +55,7 @@ if (config.doWebsocketWork && config.openSeaApiKey) {
     ],
     async (event) => {
       try {
-        if (
-          event.event_type === EventType.COLLECTION_OFFER ||
-          event.event_type === EventType.TRAIT_OFFER
-        ) {
-          logger.info("opensea-websocket", `${JSON.stringify(event)}`);
-        }
+        logger.debug("opensea-websocket", JSON.stringify(event));
```
nofir, line 58: Can you update the log to be something like:
logger.debug(
            "opensea-websocket",
            `Processing event. network=${network}, event=${JSON.stringify(event)}`
          );


I want to test creating an alarm specifically on this log and it will make it easier to ignore the other debug logs

PR #2563, opened 2022-12-07
feat: improve loading time of v5/tokens with source

jasonyjong:
## Problem
When `source` is passed in to `v5/tokens`, we return the token floor prices per token.  This leads to a more intensive query as we need to reference the `orders` table + filter on `source` in order to find the floor price by source.

## Solution
Remove expensive nested loop join query so it only needs to be run once instead of O(n), where n = # of tokens. 

See the highlighted portions below for the critical changes made

## Query
```
http://localhost:3000/tokens/v5?collection=0xbce3781ae7ca1a5e050bd9c4c77369867ebc307e&limit=20&sortBy=floorAskPrice&source=marketplace.truthlabs.co
```

## Before
Lateral join on `tst.contract = t.contract AND tst.token_id = t.token_id` means the database needs to run this join PER ROW, which is super expensive.

```
EXPLAIN ANALYZE SELECT
  t.*, s.*
FROM tokens t
JOIN LATERAL (
  SELECT o.value AS floor_sell_value
  FROM orders o
  JOIN token_sets_tokens tst ON o.token_set_id = tst.token_set_id
  WHERE
    tst.contract = t.contract AND tst.token_id = t.token_id AND o.side = 'sell' AND o.fillability_status = 'fillable' AND o.approval_status = 'approved' AND o.source_id_int = 1344
  ORDER BY o.value
  LIMIT 1
) s ON TRUE
WHERE
  (t.collection_id = '0xbce3781ae7ca1a5e050bd9c4c77369867ebc307e')
ORDER BY
  s.floor_sell_value ASC NULLS LAST,
  t.token_id
LIMIT 20;
```

Note the many ` loops=9999`
```
 Limit  (cost=1057195.65..1057195.70 rows=20 width=1108) (actual time=28422.167..28422.183 rows=20 loops=1)
   ->  Sort  (cost=1057195.65..1057208.33 rows=5070 width=1108) (actual time=28422.166..28422.180 rows=20 loops=1)
         Sort Key: o.value, t.token_id
         Sort Method: quicksort  Memory: 78kB
         ->  Nested Loop  (cost=205.14..1057060.74 rows=5070 width=1108) (actual time=367.416..28421.936 rows=27 loops=1)
               ->  Index Scan using tokens_collection_id_normalized_floor_sell_value_token_id_index on tokens t  (cost=0.69..20387.27 rows=5070 width=1101) (actual time=0.034..452.505 rows=9999 loops=1)
[...]

PR #3438, opened 2023-02-25
perf: introduce native source filter on tokens/v5

jasonyjong:
For `GET tokens/v5` endpoint

We restore the `source` filter to its behavior before [this PR](https://github.com/reservoirprotocol/indexer/pull/1726)
- This means that `source` will no longer affect the floor price/listing returned per token
- Means it should solve the perf issues faced by people passing in `source=opensea.io`

We introduce a `nativeSource` filter which behaves similarly to `source` after [this PR](https://github.com/reservoirprotocol/indexer/pull/2563)
- We add a new db query on `is_reservoir = TRUE`


jasonyjong:
@ipeleg ready for review!

ipeleg (approved):

PR #3448, opened 2023-02-28
feat: support nftx with 0x

lljxx1:


packages/contracts/test/router/v6/nftx/listings-zeroex.test.ts
```diff
@@ -0,0 +1,349 @@
...
+        expect(await nft.contract.ownerOf(nft.id)).to.eq(listings[i].vault);
+      }
+    }
+
+    // Router is stateless
+    expect(ethBalancesAfter.router).to.eq(0);
+    expect(ethBalancesAfter.nftxModule).to.eq(0);
+  };
+
+  it("Fill listing", async () =>
```
georgeroman, line 341: Can we enable all tests?

packages/contracts/test/router/v6/nftx/offers-zeroex.test.ts
```diff
@@ -0,0 +1,317 @@
...
+      }
+    }
+
+    // Router is stateless
+    expect(balancesAfter.router).to.eq(0);
+    expect(balancesAfter.seaportModule).to.eq(0);
+    expect(balancesAfter.nftxModule).to.eq(0);
+  };
+
+  it("Accpect Offer", async () => testAcceptOffers(
```
georgeroman, line 311: Can we enable all tests?

packages/indexer/yarn.lock
```diff
@@ -1,8617 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
georgeroman, line 1: I don't think we should remove this.

packages/indexer/src/sync/events/handlers/nftx.ts
```diff
@@ -449,6 +486,50 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
 
         break;
       }
+
+      case "nftx-swap-v3": {
+        const skipCheck = false;
```
georgeroman, line 491: I don't think we need this.

packages/indexer/src/sync/events/handlers/nftx.ts
```diff
@@ -449,6 +486,50 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+        const ftPool = await nftxUtils.getFtPoolDetails(baseEventParams.address, true);
```
georgeroman, line 492: This is not going to work because the ft pools are all uni-v2. What we should do instead is make an on-chain call to get `token0` and `token1` of the uni-v3 pool. Also we should have a caching layer (eg. `uniswap_v3_pools` - with a schema like `nftx_ft_pools`) so that we don't perform redundant on-chain calls.

georgeroman (changes requested):

packages/indexer/yarn.lock
lljxx1: cause of the different versions of yarn I guess?

packages/indexer/src/sync/events/handlers/nftx.ts
lljxx1: It works, uni-v3, and sushi pool both have `token0`, we could reuse the `nftx_ft_pools`.
At here we handle the uni-v3 Swap event and get the related NFTX vault. 
trigger an order to update

packages/indexer/src/sync/events/handlers/nftx.ts
lljxx1: we need to cover the uni-v3 swap to re-fetch the price data

georgeroman (changes requested):
One additional comment is that since now we support Uniswap V3 events we should also make sure to handle Uniswap V3 pools. At the moment, the `getFtPool` method only handles the old SushiSwap pool:
https://github.com/reservoirprotocol/indexer/blob/54bf81ec1857837036dc51d0f8e49b33375441f1/packages/indexer/src/utils/nftx/index.ts#L71-L94

packages/indexer/package.json
```diff
@@ -49,6 +50,7 @@
     "aws-kms-ethers-signer": "^0.1.3",
     "aws-sdk": "^2.1148.0",
     "bullmq": "1.76.1",
+    "commander": "^10.0.0",
```
georgeroman, line 53: Why do we need this? Doesn't seem to be used.

georgeroman (changes requested):

packages/indexer/src/api/endpoints/execute/get-execute-buy/v6.ts
```diff
@@ -251,6 +251,25 @@ export const getExecuteBuyV6Options: RouteOptions = {
           poolPrices[poolId].push(price);
           // Override the order's price
           order.price = price;
+
+          // Attach the ZeroEx calldata
```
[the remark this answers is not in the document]
lljxx1: @georgeroman it's handled in `router.fillBidsTx`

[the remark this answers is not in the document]
georgeroman: Discussed and let's move this logic to the sdk for consistency.

georgeroman:
@lljxx1 I think we're missing the router code for filling the updated NFTX orders via `NFTXZeroExModule` instead of the old `NFTXModule`.

packages/sdk/src/router/v6/router.ts
```diff
@@ -140,6 +140,11 @@ export class Router {
         NFTXModuleAbi,
         provider
       ),
+      nftxZeroModule: new Contract(
```
georgeroman, line 143: Let's rename this to `nftxZeroExModule`.

PR #3468, opened 2023-03-02
feat: add price/time sort directions to sales (v5)

d3or:


packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -49,6 +49,9 @@ export const getSalesV4Options: RouteOptions = {
       attributes: Joi.object()
         .unknown()
         .description("Filter to a particular attribute. Example: `attributes[Type]=Original`"),
+      sortBy: Joi.string()
+        .valid("time_asc", "time_desc", "price_asc", "price_desc")
```
ipeleg, line 53: let's split the orderBy and the sort direction to 2 different params, search for sortDirection in the code you'll see some examples

ipeleg (approved):

PR #3477, opened 2023-03-02
Deor/platf 173 add real time volume support for

d3or:


ipeleg (dismissed):
If QA OK let's sync on a time where we both around and merge it

packages/indexer/src/models/daily-volumes/daily-volume.ts
```diff
@@ -222,6 +222,84 @@ export class DailyVolume {
...
+                    ORDER BY daily_volumes.updated_at DESC LIMIT 1
+              ) ) as "volume_change"
+
+            FROM "fill_events_2" "fe"
+              JOIN "tokens" "t" ON "fe"."token_id" = "t"."token_id" AND "fe"."contract" = "t"."contract"
+              JOIN "collections" "c" ON "t"."collection_id" = "c"."id"
+            WHERE
+              "fe"."timestamp" >= $/lastDailyTimestamp/
+              AND fe.price > 0
+              AND fe.is_primary IS NOT TRUE
```
nofir, line 260: I would also add this filter:
`AND coalesce(fe.wash_trading_score, 0) = 0`

nofir (changes requested):

PR #3485, opened 2023-03-02
fix: remove is_reservoir filter from tokens endpoint

jasonyjong:
We were seeing some timeouts with this flag.

An alternative is to look into adding a better index.


jasonyjong:
@ipeleg let me know how it looks!

ipeleg (approved):

PR #3486, opened 2023-03-03
feat: use sample of highest tokens owned by user

d3or:
- feat: use sample of highest tokens owned by user
- fix: fix from clause


packages/indexer/src/api/endpoints/collections/get-user-collections/v2.ts
```diff
@@ -271,6 +278,8 @@ export const getUserCollectionsV2Options: RouteOptions = {
         ${topBidQuery}
       `;
 
+      // eslint-disable-next-line
```
ipeleg, line 281: @d3or let's remove the console.log and we can deploy this

ipeleg (approved):

PR #3514, opened 2023-03-06
fix: universe - check approvals correctly

jfrazier-eth:
Universe `checkFillability` fails due to not passing an operator address in the `isApprovedForAll` call


georgeroman:
Good catch!

PR #3529, opened 2023-03-07
feat: support seaport private order filling

lljxx1:


packages/indexer/src/api/endpoints/execute/get-execute-buy/v7.ts
```diff
@@ -371,7 +371,6 @@ export const getExecuteBuyV7Options: RouteOptions = {
                 ON orders.token_set_id = token_sets_tokens.token_set_id
               WHERE orders.id = $/id/
                 AND orders.side = 'sell'
-                AND (orders.taker = '\\x0000000000000000000000000000000000000000' OR orders.taker IS NULL)
```
georgeroman, line 374: Instead of removing this check we should have:
```sql
AND (orders.taker = '\\x0000000000000000000000000000000000000000' OR orders.taker IS NULL OR orders.taker = $/taker/)
```

packages/sdk/src/seaport-v1.4/exchange.ts
```diff
@@ -73,6 +73,43 @@ export class Exchange {
...
     if (info.side === "sell") {
+      const isEligible =
+        // Order is single quantity
+        info.amount === "1" &&
+        // Order has no criteria
+        !matchParams.criteriaResolvers &&
+        // Order requires no extra data
+        !this.requiresExtraData(order);
+
+      if (order.isPrivateOrder() && isEligible) {
```
georgeroman, line 84: I don't think we need the `isEligible` check here. We should maybe use `matchAdvancedOrder` and that will cover any types of orders.

packages/sdk/src/seaport/exchange.ts
```diff
@@ -67,15 +67,47 @@ export class Exchange {
     }
 
     if (info.side === "sell") {
-      if (
-        // Order is not private
-        recipient === AddressZero &&
+      const isEligible =
```
georgeroman, line 70: No need to duplicate all the logic for `seaport`. We can leave it for `v1.4` only.

georgeroman (changes requested):

georgeroman:
Sorry for the delay on this, let's rebase this when you get the chance.

packages/indexer/src/api/endpoints/execute/get-execute-buy/v7.ts
```diff
@@ -462,6 +462,7 @@ export const getExecuteBuyV7Options: RouteOptions = {
                 }
             `,
             {
+              taker: payload.taker,
```
georgeroman, line 465: Should wrap this in a `toBuffer` call.

packages/sdk/src/seaport-v1.4/order.ts
```diff
@@ -230,6 +235,18 @@ export class Order implements IOrder {
     }
   }
 
+  public getPrivateListingFulfillments(): Types.MatchOrdersFulfillment[] {
```
georgeroman, line 238: Can we also add this to v1.5?

georgeroman (changes requested):
Let's change the branch to `development`

PR #3538, opened 2023-03-08
fix: add backfill to correct activities collectionId

tv3636:


packages/indexer/src/jobs/backfill/backfill-activities-collection-id.ts
```diff
@@ -0,0 +1,96 @@
...
+                FROM activities a JOIN orders o ON o.id = a.order_id 
+                WHERE a.collection_id = 'collection-non-flagged'
+                AND a.type = 'bid'
+                LIMIT $/limit/
+              ) x
+              WHERE activities.id = x.activityId
+              RETURNING activities.id
+          `,
+        {
+          activityId: cursor?.activityId,
```
ipeleg, line 56: Since the update will be done till no more rows, no need for the cursor

ipeleg (changes requested):

packages/indexer/src/jobs/backfill/backfill-activities-collection-id.ts
tv3636: removed the cursor

ipeleg (approved):

PR #3544, opened 2023-03-08
fix: unicode in header

xcaptain:
In some cases, `new Date(Date.now()).toString()` will result unicode, which causes http response fail. I recommend using a unix timestamp here to avoid unicode.

```
> new Date(Date.now()).toString()
'Wed Mar 08 2023 18:31:12 GMT+0800 (中国标准时间)'
```


xcaptain:
@georgeroman can you take a look at it?

georgeroman:
@xcaptain Nice catch! However, since this is a breaking change we're going to skip it for the moment. It's not impacting our production instances.

PR #3550, opened 2023-03-08
feat: added includeDynamicPricing flag to user tokens endpoint

JadenDurnford:
Angel said that testing for this story is difficult on goerli and that he would verify the changes once it is on prod, I did some testing locally and the flag was working as intended.


nofir (approved):

nofir (changes requested):
Similar to the comments in https://github.com/reservoirprotocol/indexer/pull/3599. Given that it seems like we are adding in more apis, we should make this more generic.

nofir (approved):

PR #3553, opened 2023-03-08
feat: add floor ask to top bid event

d3or:


packages/indexer/src/jobs/websocket-events/events/new-top-bid-websocket-event.ts
```diff
@@ -98,6 +121,32 @@ export class NewTopBidWebsocketEvent {
             fromBuffer(order.currency)
           ),
           criteria: order.criteria,
+          floorAsk: {
```
nofir, line 124: A few comments:
1. i think this needs to be under collection property as its the collection floor ask.
2. I do believe we should just have collection -> floorAskPrice (maybe the id too) / or just collectionFloorAskPrice as this is probably to better filter certain bids rather for display purposed (the criteria obj also include already the image, description, etc). with websocket events, its much harder to remove information than to add something later where with apis you can just just create a new version and deprecate a property)
3. lets see what peter says regarding which value to use (or include all)

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/events/new-top-bid-websocket-event.ts
```diff
@@ -71,35 +119,24 @@ export class NewTopBidWebsocketEvent {
...
+          },
+
           criteria: order.criteria,
         },
         owners: ownersChunk,
+        colllection: {
+          id: order.collection_id,
+          slug: order.collection_slug,
+          name: order.collection_name,
+          floorAskPrice: order.floor_sell_value,
```
nofir, line 136: Needs formatEth (same for floorAskPriceNormalized)

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/events/new-top-bid-websocket-event.ts
```diff
@@ -71,35 +119,24 @@ export class NewTopBidWebsocketEvent {
...
+            currency: price.currency,
+            amount: price.amount,
+            netAmount: price.netAmount,
+            normalizedNetAmount: priceNormalized.netAmount,
+          },
+
           criteria: order.criteria,
         },
         owners: ownersChunk,
+        colllection: {
```
nofir, line 132: looks like spelling mistake, should be "collection"

nofir (changes requested):

nofir (approved):

PR #3555, opened 2023-03-09
Tv/platf 840 merge lastBuy and lastSell to includeLastSale with royaltyBreakdown

tv3636:


nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v7.ts
```diff
@@ -0,0 +1,595 @@
...
+              name: r.collection_name,
+              imageUrl: r.metadata?.imageUrl,
+              floorAskPrice: r.collection_floor_sell_value
+                ? formatEth(r.collection_floor_sell_value)
+                : null,
+            },
+            lastSale: query.includeLastSale
+              ? {
+                  value:
+                    r.last_buy_timestamp > r.last_sell_timestamp
```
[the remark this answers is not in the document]
tv3636: updated to include the full lastSale, mirroring topBid

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v7.ts
```diff
@@ -0,0 +1,627 @@
...
+      selectRoyaltyBreakdown = ", r.*";
+      includeRoyaltyBreakdownQuery = `
+        LEFT JOIN LATERAL (
+        SELECT
+            CASE WHEN fe.royalty_fee_breakdown IS NOT NULL and jsonb_array_length(fe.royalty_fee_breakdown) > 0 THEN fe.royalty_fee_breakdown::json->0
+            WHEN o.fee_breakdown IS NULL THEN '{}'::json 
+            WHEN 'royalty' IN (SELECT jsonb_array_elements(o.fee_breakdown)->>'kind') THEN o.fee_breakdown::json->1
+            ELSE '{}'::json END AS royalty_breakdown,
+            fe.timestamp AS last_sale_time,          
+            o.id AS last_sale_id,
```
nofir, line 305: Remind me, do we actually need the join with the orders table? don't you have all the info already in the sales table? Also, i see up until now lastBuy / lastSell only exposed value, timestamp, is the requirement to expose id, maker, source, etc?

tv3636: fill_events2 does not have `value` or `currency_value` which are both used to fill out the `JoiPrice` object (at least in other APIs), maker I included as I was going off of `topBid` as an example. 

initially the reason for joining orders and fill_events2 was to get the most complete royalty breakdown possible, since neither table has the complete data yet

nofir (commented):
Looks good. One question, Why is there a large diff in v6?

nofir (approved):

nofir (approved):

PR #3558, opened 2023-03-09
feat: add sorting by time/price to sales v4

d3or:


packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -224,6 +232,18 @@ export const getSalesV4Options: RouteOptions = {
       query.endTimestamp = 9999999999;
     }
 
+    let orderDirection = "DESC";
+    if (query.sortDirection === "asc") {
+      orderDirection = "ASC";
+    }
```
ipeleg, line 238: Can be removed as sort direction will be either asc or desc, in the query we can add  $/orderDirection/ and pass the query as we already do

packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -224,6 +232,18 @@ export const getSalesV4Options: RouteOptions = {
...
+
+    // Default to ordering by time
+    let queryOrderBy = `ORDER BY fill_events_2.timestamp ${orderDirection}, fill_events_2.log_index ${orderDirection}, fill_events_2.batch_index ${orderDirection}`;
+
+    if (query.orderBy && query.orderBy === "price") {
```
ipeleg, line 243: I think we'll need an index to support order by price

ipeleg (approved):

ipeleg (approved):

PR #3585, opened 2023-03-10
feat: royalties improve

lljxx1:


packages/indexer/src/sync/events/handlers/royalties/core.ts
```diff
@@ -149,6 +155,8 @@ export async function extractRoyalties(
 
   shouldExcludeAddressList.add(Sdk.Common.Addresses.Weth[config.chainId]);
   shouldExcludeAddressList.add(Sdk.Common.Addresses.Eth[config.chainId]);
+  // bendWETH
+  shouldExcludeAddressList.add("0xed1840223484483c0cb050e6fc344d1ebf0778a9");
```
georgeroman, line 159: Let's add this to the `bend-dao` logic in the core sdk.

packages/contracts/contracts/router/modules/Permit2Module.sol
```diff
@@ -20,6 +21,9 @@ contract Permit2Module is BaseModule {
     IAllowanceTransfer.AllowanceTransferDetails[] calldata transferDetails,
     bytes calldata signature
   ) external nonReentrant {
+    if (owner != tx.origin) {
+      revert Unauthorized();
```
georgeroman, line 25: We should also define the error above.

PR #3598, opened 2023-03-10
feat: adding sortBy updated_at to orders/asks/v4

JadenDurnford:


packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -418,6 +424,8 @@ export const getOrdersAsksV4Options: RouteOptions = {
         } else {
           baseQuery += ` ORDER BY orders.price, orders.id`;
         }
+      } else if (query.sortBy === "updatedAt") {
+        baseQuery += ` ORDER BY orders.updated_at DESC, orders.id DESC`;
```
nofir, line 428: - I am not sure DESC is the right approach for a firehose if the approach is to catch up from a certain point in time.
- Line 267, I think you need to remove the restriction for only "active" asks in order to get updates.
- For now, I would disable filters when sorting by updated at.

nofir (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -227,7 +228,6 @@ export const getOrdersAsksV4Options: RouteOptions = {
               ELSE 'active'
             END
           ) AS status,
-          orders.updated_at,
           (${criteriaBuildQuery}) AS criteria
           ${query.includeRawData ? ", orders.raw_data" : ""}
         FROM orders
```
nofir, line 233: see line 250, you also need to change that to be based on the updated_at when sorting by updatedAt

nofir (changes requested):

PR #3599, opened 2023-03-10
feat: adding dynamic pricing info to orders/asks/v4

JadenDurnford:


packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -229,7 +237,7 @@ export const getOrdersAsksV4Options: RouteOptions = {
           ) AS status,
           orders.updated_at,
           (${criteriaBuildQuery}) AS criteria
-          ${query.includeRawData ? ", orders.raw_data" : ""}
+          ${query.includeRawData || query.includeDynamicPricing ? ", orders.raw_data" : ""}
```
nofir, line 240: Why is that needed?

packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -481,6 +489,79 @@ export const getOrdersAsksV4Options: RouteOptions = {
           source = sources.get(Number(r.source_id_int));
         }
 
+        // Use default currencies for backwards compatibility with entries
+        // that don't have the currencies cached in the tokens table
+        const floorAskCurrency = r.currency
+          ? fromBuffer(r.currency)
```
nofir, line 495: In general, we want to start having more standardized responses and avoid duplicating logic between different apis. Unless there is a strong case , i would create a `DynamicPricingJoi` object and `getJoiDynamicPricingObject` to populate it (similar to how we do with price across apis)

packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -481,6 +489,79 @@ export const getOrdersAsksV4Options: RouteOptions = {
...
+        let dynamicPricing = undefined;
+        if (query.includeDynamicPricing) {
+          // Add missing royalties on top of the raw prices
+          const missingRoyalties = query.normalizeRoyalties
+            ? ((r.missing_royalties ?? []) as any[])
+                .map((mr: any) => bn(mr.amount))
+                .reduce((a, b) => a.add(b), bn(0))
+            : bn(0);
+
+          if (r.dynamic && r.kind === "seaport") {
```
nofir, line 507: What about support for seaport 1.4?

nofir (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
nofir: @georgeroman im assuming this is something we need to add in all endpoints that have the dynamic pricing logic

nofir (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
georgeroman: Right

nofir (approved):
Looks good! two comments:
- can you update to `if (dynamic && (kind === "seaport" || kind === "seaport-v1.4")) {` in other endpoints? or at least create another pr for it?
- Are there any other endpoints that can benefit from the JoiDynamicPrice obj? (i know you have another pr with similar changes.

JadenDurnford:
Yup, I'll create a ticket for it and add the logic in for the other endpoints

PR #3601, opened 2023-03-10
Add Notes for "Create Collection Set"

jamiechicago312:
In `notes` line 12, I added a description and example for users to understand `collectionSetId` better in the docs.


nofir (changes requested):

jamiechicago312:
I changed it, let me know if I need anything else 👍

nofir (approved):

PR #3623, opened 2023-03-14
feat: add JoiSaleObject

tv3636:
FYI @nofir here's a WIP on adding a `JoiSaleObject` - this is functional but wanted to confirm this is along the lines of what we'd want

### DIFF
--- a/packages/indexer/src/common/joi.ts
+++ b/packages/indexer/src/common/joi.ts
@@ -190,9 +190,10 @@ export const JoiOrderCriteria = Joi.alternatives(
 
 // --- Sales ---
 
-const JoiFeeBreakdown = Joi.object({
+export const JoiFeeBreakdown = Joi.object({
   kind: Joi.string(),
-  detail: Joi.object().optional(),
+  bps: Joi.number(),
+  recipient: Joi.string(),
 });
 
 export const JoiSale = Joi.object({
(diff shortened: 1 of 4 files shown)


packages/indexer/src/common/joi.ts
```diff
@@ -185,3 +187,106 @@ export const JoiOrderCriteria = Joi.alternatives(
...
+export const JoiSale = Joi.object({
+  price: JoiPrice,
+  timestamp: Joi.number(),
+  royaltyFeeBps: Joi.number().optional(),
+  marketplaceFeeBps: Joi.number().optional(),
+  paidFullRoyalty: Joi.boolean().optional(),
+  feeBreakdown: Joi.array().items(JoiFeeBreakdown).optional(),
+});
+
+export const getJoiSaleObject = async (
```
nofir, line 207: Does it make sense to change it to getJoinLastSaleObject?

packages/indexer/src/common/joi.ts
```diff
@@ -185,3 +187,106 @@ export const JoiOrderCriteria = Joi.alternatives(
...
+    marketplaceFeeBps?: number;
+    paidFullRoyalty?: boolean;
+    royaltyFeeBreakdown?: any;
+    marketplaceFeeBreakdown?: any;
+  },
+  currencyAddress: string,
+  timestamp: number
+) => {
+  const currency = await getCurrency(currencyAddress);
+  const lastSaleFeeInfoIsValid = (fees.royaltyFeeBps ?? 0) + (fees.marketplaceFeeBps ?? 0) < 10000;
```
nofir, line 232: Is there any way to also utilize the changes for the logic in the sales api?
https://github.com/reservoirprotocol/indexer/blob/71964c1808b81abc0a8241e4917bbd2ecf8d482d/packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts#L331

maybe have another method that just gets the fee info and then utilized both in sales and the apis you changed.

PR #3646, opened 2023-03-16
feat: bulk cancel

lljxx1:


packages/indexer/src/api/routes.ts
```diff
@@ -643,6 +643,12 @@ export const setupRoutes = (server: Server) => {
     options: executeEndpoints.getExecuteCancelV2Options,
   });
 
+  server.route({
+    method: "GET",
```
georgeroman, line 647: We should make this a `POST` request

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,418 @@
...
+
+import { redb } from "@/common/db";
+import { logger } from "@/common/logger";
+import { bn, fromBuffer, regex } from "@/common/utils";
+import { config } from "@/config/index";
+
+const version = "v3";
+
+export const getExecuteCancelV3Options: RouteOptions = {
+  description: "Cancel order",
```
georgeroman, line 17: Let's update the description and notes to:
- description: `Cancel orders`
- notes: `Cancel existing orders on any marketplace`

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,418 @@
...
+        Joi.object({
+          id: Joi.alternatives()
+            .try(Joi.array().items(Joi.string()).max(20), Joi.string())
+            .description(
+              "Array of Order IDs. Example: `0x8d04a8c79ceb0889bdd12acdf3fa9d207ed3ff63`"
+            )
+            .optional(),
+        })
+      ),
+      softCancel: Joi.boolean()
```
georgeroman, line 47: We can remove the `softCancel` option in this API version.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,418 @@
...
+  description: "Cancel order",
+  notes: "Cancel an existing order on any marketplace",
+  tags: ["api", "Router"],
+  plugins: {
+    "hapi-swagger": {
+      order: 11,
+    },
+  },
+  validate: {
+    payload: Joi.object({
```
georgeroman, line 26: Let's have the following payload format (also see the comments of what we want to support in this initial release and what we can skip):
```javascript
      Joi.alternatives(
        // If all order ids have the kind seaport or seaport-v1.4 then cancel them all
        // Otherwise throw an error (we don't support multi-order cancelling for other order kinds)
        Joi.object({
          kind: "orderIds",
          data: {
            orderIds: Joi.array().items(Joi.string()).min(1).required(),
          },
        }),
        // If the order kind is not seaport or seaport-v1.4 throw an error
        Joi.object({
          kind: "token",
          data: {
            side: "sell",
            orderKind: Joi.string().required(),
            token: Joi.string().pattern(regex.token).required(),
          },
        }),
        // If the order kind is not seaport or seaport-v1.4 throw an error
        Joi.object({
          kind: "maker",
          data: {
            orderKind: Joi.string().required(),
            maker: Joi.string().pattern(regex.address).required(),
          },
        })
      );
```

georgeroman (changes requested):

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+  notes: "Cancel existing orders on any marketplace",
+  tags: ["api", "Router"],
+  plugins: {
+    "hapi-swagger": {
+      order: 11,
+    },
+  },
+  validate: {
+    payload: Joi.object({
+      params: Joi.alternatives().try(
```
georgeroman, line 27: No need to wrap in another `params` object, the alternatives can just be in the top-level (eg. `payload: Joi.alternatives()`.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+        // If all order ids have the kind seaport or seaport-v1.4 then cancel them all
```
georgeroman, line 28: These comments were just to make it clear for you what we want to support and what we don't. We should remove them both here and below.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+  handler: async (request: Request) => {
+    const payload = request.payload as any;
+
+    const params = payload.params;
+    const actionData = params.data;
+
+    // Cancel by maker
+    if (params.kind === "maker") {
+      if (!actionData.maker) {
+        throw Boom.badData("maker must provide");
```
georgeroman, line 95: No need to check for the presence of `maker`, the joi validation will ensure that it is present.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+                },
+                orderIndex: 0,
+              },
+            ],
+          },
+        ],
+      };
+    }
+
+    // Cancel by tokenId or multile orderIds
```
georgeroman, line 147: Typp, should be `multiple`

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+    try {
+      // Fetch the order to get cancelled
```
georgeroman, line 149: Should be `Fetch the orders`, since we can have multiple orders.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+            AND (orders.fillability_status = 'fillable' OR orders.fillability_status = 'no-balance')
+        `,
+              {
+                token_set_id: `token:${actionData.token}`,
+                order_kind: actionData.orderKind,
+              }
+            )
+          : await redb.manyOrNone(
+              `
+          SELECT
```
georgeroman, line 171: We should properly indent this.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+
+      let cancelTx: TxData;
+      let orderSide: "sell" | "buy";
+
+      const maker = fromBuffer(orderResult.maker);
+
+      // REFACTOR: Move to SDK and handle X2Y2
+      switch (orderResult.kind) {
+        case "seaport": {
+          if (isBulkCancel) {
```
georgeroman, line 212: For simplicity, let's aways use the `cancelOrdersTx` method, this way we don't need the `if (isBulkCancel)` line anymore.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+
+            cancelTx = exchange.cancelOrderTx(maker, order);
+            orderSide = order.getInfo()!.side;
+          }
+
+          break;
+        }
+
+        case "seaport-v1.4": {
+          if (isBulkCancel) {
```
georgeroman, line 232: Same comment here as above regarding `isBulkCancel`.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        // TODO: Add support for X2Y2 (it's tricky because of the signature requirement)
+
+        default: {
+          throw Boom.notImplemented("Unsupported order kind");
+        }
+      }
+
+      // TODO: We should remove the "listing"/"offer" distinction once we get to bundles
```
georgeroman, line 319: We can remove this comment since it's not relevant anymore.

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -0,0 +1,352 @@
...
+      return {
+        steps: [
+          {
+            id: "cancellation",
+            action: orderSide === "sell" || isBulkCancel ? "Submit cancellation" : "Cancel offer",
```
georgeroman, line 324: The action should always be `Submit cancellation`.

georgeroman (changes requested):

packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -23,40 +23,47 @@ export const getExecuteCancelV3Options: RouteOptions = {
...
-        })
-      ),
-      maxFeePerGas: Joi.string()
-        .pattern(regex.number)
-        .description("Optional. Set custom gas price"),
-      maxPriorityFeePerGas: Joi.string()
-        .pattern(regex.number)
-        .description("Optional. Set custom gas price"),
-    }),
+    payload: Joi.alternatives().try(
```
georgeroman, line 26: My bad here, sorry, I thought we can easily share the common parameters `maxFeePerGas` and `maxPriorityFeePerGas` but looks like we need to duplicate them. Let's go back to your initial implementation with the `params` field.

PR #3687, opened 2023-03-20
Feat: Searching for an array of tokens in asks

Tevzi2:
Implements the ability to search for an array of tokens instead of a single token in orders/asks, proposition is implemented as v5 of orders asks.

### DIFF
--- a/packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
+++ b/packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
@@ -1,0 +1,571 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import * as Boom from "@hapi/boom";
+import { Request, RouteOptions } from "@hapi/hapi";
+import * as Sdk from "@reservoir0x/sdk";
+import Joi from "joi";
+import _ from "lodash";
+
+import { redb } from "@/common/db";
+import { logger } from "@/common/logger";
+import {
+  JoiPrice,
+  getJoiPriceObject,
+  JoiOrderCriteria,
+  JoiDynamicPrice,
+  getJoiDynamicPricingObject,
+} from "@/common/joi";
+import {
+  buildContinuation,
+  fromBuffer,
+  getNetAmount,
+  regex,
+  splitContinuation,
+  toBuffer,
+} from "@/common/utils";
+import { config } from "@/config/index";
+import { CollectionSets } from "@/models/collection-sets";
+import { Sources } from "@/models/sources";
+import { SourcesEntity } from "@/models/sources/sources-entity";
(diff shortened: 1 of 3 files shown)


packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
```diff
@@ -0,0 +1,571 @@
...
+  },
+  validate: {
+    query: Joi.object({
+      ids: Joi.alternatives(Joi.array().items(Joi.string()), Joi.string()).description(
+        "Order id(s) to search for."
+      ),
+      tokens: Joi.alternatives()
+        .try(
+          Joi.array().items(Joi.string().lowercase().pattern(regex.token)),
+          Joi.string().lowercase().pattern(regex.address)
```
ipeleg, line 52: should be regex.token, same as the pattern for the array items

packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
```diff
@@ -0,0 +1,571 @@
...
+          conditions.push(`orders.id IN ($/ids:csv/)`);
+        } else {
+          conditions.push(`orders.id = $/ids/`);
+        }
+      } else {
+        orderStatusFilter = `orders.fillability_status = 'fillable' AND orders.approval_status = 'approved'`;
+      }
+
+      if (query.tokens) {
+        if(!_.isArray(query.tokens)) {
```
ipeleg, line 283: Shouldn't this if check be `if(_.isArray(query.tokens))` ? if not the map inside the if will fail as it's not an array according to the current check

packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
```diff
@@ -0,0 +1,571 @@
...
+          (query as any).tokensPrefix = [query.tokens].map(token => `token:${token}`);
+        }
+
+        (query as any).tokensFilter = query.tokensPrefix.map(toBuffer);
```
ipeleg, line 287: `query.tokensPrefix` can be undefined hence the .map will fail

ipeleg, line 287: no need to to toBuffer as token_set_id in the orders table is a string and not a byte type column

ipeleg (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
Tevzi2: My bad wrote what happens inside of the if statement incorrectly, if tokens is not an array => convert it to an array, then map

PR #3692, opened 2023-03-20
Tv/platf 831 standardize price obj in apis

tv3636:


packages/indexer/src/models/activities/index.ts
```diff
@@ -572,6 +581,8 @@ export class Activities {
                     source_id_int AS "order_source_id_int",
                     side AS "order_side",
                     kind AS "order_kind",
+                    currency AS "order_currency",
```
nofir, line 584: see above

nofir (changes requested):

PR #3695, opened 2023-03-20
Feat: Searching for an array of tokens in bids

Tevzi2:
Implements the ability to search for an array of tokens instead of a single token in orders/bids, proposition is implemented in v5 of orders bids.

### DIFF
--- a/packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
+++ b/packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
@@ -34,11 +34,13 @@ export const getOrdersBidsV5Options: RouteOptions = {
       ids: Joi.alternatives(Joi.array().items(Joi.string()), Joi.string()).description(
         "Order id(s) to search for (only fillable and approved orders will be returned)"
       ),
-      token: Joi.string()
-        .lowercase()
-        .pattern(regex.token)
+      tokens: Joi.alternatives()
+        .try(
+          Joi.array().items(Joi.string().lowercase().pattern(regex.token)),
+          Joi.string().lowercase().pattern(regex.address)
+        )
         .description(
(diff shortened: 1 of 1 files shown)


packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
```diff
@@ -34,11 +34,13 @@ export const getOrdersBidsV5Options: RouteOptions = {
       ids: Joi.alternatives(Joi.array().items(Joi.string()), Joi.string()).description(
         "Order id(s) to search for (only fillable and approved orders will be returned)"
       ),
-      token: Joi.string()
-        .lowercase()
-        .pattern(regex.token)
+      tokens: Joi.alternatives()
+        .try(
+          Joi.array().items(Joi.string().lowercase().pattern(regex.token)),
+          Joi.string().lowercase().pattern(regex.address)
```
ipeleg, line 40: Need to be regex.token to match the pattern in the array

packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
```diff
@@ -34,11 +34,13 @@ export const getOrdersBidsV5Options: RouteOptions = {
       ids: Joi.alternatives(Joi.array().items(Joi.string()), Joi.string()).description(
         "Order id(s) to search for (only fillable and approved orders will be returned)"
       ),
-      token: Joi.string()
-        .lowercase()
-        .pattern(regex.token)
+      tokens: Joi.alternatives()
+        .try(
+          Joi.array().items(Joi.string().lowercase().pattern(regex.token)),
```
ipeleg, line 39: Let's add a limit on the size of the array to be on the safe side, I think up to 100 tokens should be enough

ipeleg (changes requested):

PR #3696, opened 2023-03-20
feat: adding index for updated at ascending for asks endpoint

JadenDurnford:


packages/indexer/src/migrations/1643809164129_orders.sql
```diff
@@ -131,6 +131,10 @@ ALTER TABLE "orders" SET (
   autovacuum_analyze_threshold=100000
 );
 
+-- CREATE INDEX "orders_asks_updated_at_asc_id_index"
```
ipeleg, line 134: What is this for? and i guess if we want it we want to be uncommented right?

JadenDurnford: We created this index for the lightnode, @nofir should this be uncommented now that it has been successfully created?

nofir (dismissed):

packages/indexer/src/migrations/1643809164129_orders.sql
ipeleg, line 134: @JadenDurnford as it's now in a migration file that was already executed we can uncomment this

ipeleg (changes requested):

packages/indexer/src/migrations/1643809164129_orders.sql
nofir: @JadenDurnford ust notice that we will have to run it manually in polygon, etc

nofir (approved):

ipeleg (approved):

PR #3702, opened 2023-03-21
feat: marketplace blacklist

lljxx1:


georgeroman (changes requested):

packages/indexer/src/orderbook/orders/blur/index.ts
```diff
@@ -63,6 +64,14 @@ export const save = async (
         });
       }
 
+      const isFiltered = await checkMarketplaceIsFiltered(order.params.collection, "blur");
```
georgeroman, line 67: Just to make sure nothing breaks when we deploy this, let's comment this out (applies to the other exchanges where you added this check). We can then trigger a refresh for a few collections and make sure the blacklist is properly updated. Once everything looks good, we can uncomment this.

packages/indexer/src/migrations/1680055939007_contracts_marketplace-blacklist.sql
```diff
@@ -0,0 +1,7 @@
+-- Up Migration
+
+ALTER TABLE "contracts" ADD COLUMN "marketplace_blacklists" JSONB;
```
georgeroman, line 3: Let's keep the naming consistent with what OS has and call this `filtered_operators`.

packages/indexer/src/migrations/1680055939007_contracts_marketplace-blacklist.sql
```diff
@@ -0,0 +1,7 @@
+-- Up Migration
```
georgeroman, line 1: Can you also change the name of the migration file to `contracts-filtered-operators`?

packages/sdk/src/seaport-v1.1/addresses.ts
```diff
@@ -28,3 +28,14 @@ export const OpenseaConduitKey: ChainIdToAddress = {
   [Network.Polygon]: "0x0000007b02230091a7ed01230072f7006a004d60a8d4e71d599b8104250f0000",
   [Network.Optimism]: "0x0000007b02230091a7ed01230072f7006a004d60a8d4e71d599b8104250f0000",
 };
+
+export const OpenseaConduit: ChainIdToAddress = {
```
georgeroman, line 32: We don't store the conduit anymore, but use the `deriveConduit` method to derive the address from the key.

packages/indexer/src/utils/marketplace-blacklists/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  {
+    address: Sdk.LooksRare.Addresses.TransferManagerErc1155[config.chainId],
+    marketplace: "looks-rare",
+  },
+  {
+    address: Sdk.Nftx.Addresses.MarketplaceZap[config.chainId],
+    marketplace: "nftx",
+  },
+  {
+    address: Sdk.Seaport.Addresses.OpenseaConduit[config.chainId],
```
georgeroman, line 32: I don't think this will compile, it's called `SeaportV11` now.

georgeroman (changes requested):

PR #3705, opened 2023-03-21
fix: refactor bulk cancel params

lljxx1:


packages/indexer/src/api/endpoints/execute/get-execute-cancel/v3.ts
```diff
@@ -24,36 +24,32 @@ export const getExecuteCancelV3Options: RouteOptions = {
...
+      token: Joi.string().pattern(regex.token).optional(),
       maxFeePerGas: Joi.string()
         .pattern(regex.number)
         .description("Optional. Set custom gas price"),
       maxPriorityFeePerGas: Joi.string()
         .pattern(regex.number)
         .description("Optional. Set custom gas price"),
-    }),
+    })
+      .oxor("orderIds")
```
georgeroman, line 50: Can we use something like this instead: https://github.com/reservoirprotocol/indexer/blob/a078eda3164ade9eb1ef42dc532f0221abb1873e/packages/indexer/src/api/endpoints/execute/get-execute-buy/v7.ts#L80-L81

This will ensure that only one of `token`, `maker` and `orderIds` will be passed.

PR #3707, opened 2023-03-21
Enable commitlint

fdmota:
Ensure that commits follow the right message format which is important for later versioning.


package-lock.json
```diff
@@ -12,8 +12,6 @@
         "packages/*"
       ],
       "dependencies": {
-        "@commitlint/cli": "^16.2.1",
```
georgeroman, line 15: Can we remove `package-lock.json`? The convention is to use `yarn` instead of `npm` and so the `npm` lock file is redundant.

fdmota: I can, but the package-lock.json already exists in the repo. Should I remove it?

georgeroman: I think so, nothing should depend on that. @nofir @ipeleg do you agree?

nofir: yeah, i think it can be removed.

fdmota: removed

georgeroman (approved):

PR #3716, opened 2023-03-21
Tv/platf 890 cleanup empty collections after being

tv3636:


packages/indexer/src/jobs/metadata-index/write-queue.ts
```diff
@@ -132,6 +132,30 @@ if (config.doBackgroundWork) {
             `${contract}:${tokenId}`
           );
 
+          // Update the token_count on both the new and old collection
+          for (const collectionId of [collection, result.collectionId]) {
```
nofir, line 136: We already do the logic to calc the token count inside fetchCollectionMetadata (for the new collection).

nofir: There is also a job to recalc the token count that can be used:
https://github.com/reservoirprotocol/indexer/blob/049187fb2d47e906d460a559ba91352ab49739b0/packages/indexer/src/jobs/token-updates/fetch-collection-metadata.ts#L128

nofir (approved):

PR #3727, opened 2023-03-22
fix: leave includeRoyaltiesPaid undefined if token has no last order

tv3636:
`r.royalties_paid` can return NULL if there is no matching order ID for the token, so leave `includeRoyaltiesPaid` undefined if there is no order in the first place

### DIFF
--- a/packages/indexer/src/api/endpoints/tokens/get-tokens/v5.ts
+++ b/packages/indexer/src/api/endpoints/tokens/get-tokens/v5.ts
@@ -1154,7 +1154,10 @@ export const getTokensV5Options: RouteOptions = {
                   feeBreakdown: feeBreakdown,
                 }
               : undefined,
-            royaltiesPaid: query.includeRoyaltiesPaid ? r.royalties_paid : undefined,
+            royaltiesPaid:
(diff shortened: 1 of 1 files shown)


packages/indexer/src/api/endpoints/tokens/get-tokens/v5.ts
```diff
@@ -1154,7 +1154,10 @@ export const getTokensV5Options: RouteOptions = {
                   feeBreakdown: feeBreakdown,
                 }
               : undefined,
-            royaltiesPaid: query.includeRoyaltiesPaid ? r.royalties_paid : undefined,
+            royaltiesPaid:
+              query.includeRoyaltiesPaid && r.royalties_paid !== null
```
nofir, line 1158: is royalties_paid null or undefined? you might need **!=** instead

tv3636: royalties_paid can be null, so this is meant to check explicitly for null. if it was undefined then it would simply return undefined which is expected too (although it should never be undefined if it reaches this check)

nofir (approved):

PR #3828, opened 2023-03-28
fix: update job id for recalculate old collection token count

tv3636:


packages/indexer/src/jobs/collection-updates/recalc-token-count-queue.ts
```diff
@@ -50,6 +50,14 @@ if (config.doBackgroundWork) {
   });
 }
 
-export const addToQueue = async (collection: string, delay = 60 * 1000) => {
-  await queue.add(collection, { collection }, { delay, jobId: collection });
+export const addToQueue = async (
```
nofir, line 53:
```
export const addToQueue = async (
  collection: string,
  delay = 60 * 1000,
  force = false
) => {
  await queue.add(
    collection,
    { collection },
    { delay, jobId: force ? undefined : collection }
  );
};
```

nofir: I think something like that would be simpler (also, in your changes, you didnt actually change the jobId)

nofir (changes requested):

packages/indexer/src/jobs/collection-updates/recalc-token-count-queue.ts
tv3636: ah oops was looking at the job name. updated

packages/indexer/src/jobs/collection-updates/recalc-token-count-queue.ts
```diff
@@ -50,6 +50,6 @@ if (config.doBackgroundWork) {
   });
 }
 
-export const addToQueue = async (collection: string, delay = 60 * 1000) => {
-  await queue.add(collection, { collection }, { delay, jobId: collection });
+export const addToQueue = async (collection: string, force = false, delay = 60 * 1000) => {
```
nofir, line 53: is there any place in the code that is currently calling the addToQueue with delay? otherwise, adding the force before the delay might break it

nofir (approved):

PR #3829, opened 2023-03-28
Tv/platf 673 add query payloads to datadog traces

tv3636:


packages/indexer/src/models/api-keys/index.ts
```diff
@@ -210,6 +212,20 @@ export class ApiKeyManager {
   public static async logRequest(request: Request) {
     const log: any = await ApiKeyManager.getBaseLog(request);
     logger.info("metrics", JSON.stringify(log));
+
+    // Add request params to Datadog trace
+    try {
+      const requestParams: any = flat.flatten({ ...log.payload, ...log.query, ...log.params });
```
nofir, line 218: Any reason to flatten it and not keep payload, query, params? this is how the metrics log works too.

packages/indexer/src/models/api-keys/index.ts
```diff
@@ -210,6 +212,20 @@ export class ApiKeyManager {
...
+      Object.keys(requestParams).forEach(
```
nofir, line 219: Why is that needed?

packages/indexer/src/models/api-keys/index.ts
tv3636: we need to flatten them because Datadog doesn't support nested objects in traces. 

we could keep them separate, but for convenience in searching across traces i think we'd rather have them all in one (this way you can just filter on contract and immediately get all traces regardless of where the contract was specified)

packages/indexer/src/models/api-keys/index.ts
tv3636: before casting to string there were a bunch of traces dropped. Datadog only supports strings, numbers, or (non-nested) objects in the trace value

ipeleg (approved):

PR #3835, opened 2023-03-29
Optimize seaport exchange

xcaptain:
Currently I found seaport/exchange.ts and seaport-v1.4/exchanges have a lot of code in common
image
I want to extract the common code into seaport-base then only a few code left in seaport and seaport-v1.4
image

At first I was mean to only abstract exchange, but later I found I have to also change the order and builder to make tests work, so this PR is becoming very big, to make review easier I will list what I have done here.

1. Move `builder` folder to `seaport-base`, `seaport/builder` and `seaport-v1.4/builder` are identic
2. Added a `IOrder` interface, so exchange depends on an interface rather than a concrete order
3. Added a `SeaportBaseExchange`, `SeaportExchange` and `SeaportV14Exchange` will inherit`SeaportBaseExchange`
4. Copy `types.ts` to `seaport-base` because `seaport/types.ts` and `seaport-v1.4/types.ts` are identic
5. Update `BaseBuilder` class, added a generic `build` method to build concrete order
6. Fixes some tests to use the new generic `build` method
7. Remove unused code in sdk
8. Update indexer to make it build

Status:
The whole workspace builds, contract tests passed
image

Need to do more tests


packages/sdk/src/seaport-base/types.ts
```diff
@@ -11,6 +11,7 @@ export enum ItemType {
   ERC1155_WITH_CRITERIA,
 }
 
+// TODO: @Joey keep contract?
```
xcaptain, line 14: `SeaportV11.OrderType` has a little different with `SeaportV14.OrderType`, can we share a type here? I think it's ok, so I will remove this TODO tag.

packages/sdk/src/seaport-base/types.ts
```diff
@@ -84,6 +85,7 @@ export type OrderComponents = {
   signature?: string;
 };
 
+// TODO: @Joey BigNumberish or string?
 export type MatchParams = {
```
xcaptain, line 89: `BigNumberish ` is a union type of `string`, should be safe to use `BigNumberish ` here.

PR #3848, opened 2023-03-29
feat: update 1day volume to handle collections with no sales in 24h

d3or:


packages/indexer/src/models/daily-volumes/daily-volume.ts
```diff
@@ -304,6 +312,40 @@ export class DailyVolume {
...
+      (c: any) => !updatedCollectionIds.includes(c.id)
+    );
+
+    if (collectionsToUpdateToNull.length) {
+      const updateToNullQueries = collectionsToUpdateToNull.map((c: any) => {
+        return {
+          query: `
+          UPDATE collections
+          SET
+            day1_volume = NULL,
```
nofir, line 326: might make sense to default day1_volume to 0 (this is the default in the db)

packages/indexer/src/models/daily-volumes/daily-volume.ts
```diff
@@ -230,6 +230,14 @@ export class DailyVolume {
     const currentDate = new Date();
     const startTime = new Date(currentDate.getTime() - 24 * 60 * 60 * 1000).getTime() / 1000;
 
+    // Get a list of all collections that have non-null 1day values
+    const collectionsWith1DayValues = await ridb.manyOrNone(
```
nofir, line 234: would it be possible to actually update it in one query in postgres instead of in the code? not critical.

nofir (dismissed):

nofir (approved):

PR #3875, opened 2023-03-31
Optimize seaport orderbook1

xcaptain:


packages/sdk/src/seaport-base/order.ts
```diff
@@ -2,12 +2,19 @@ import * as Types from "./types";
 import { BaseOrderInfo } from "./builders/base";
 import { BigNumberish } from "@ethersproject/bignumber";
 
+export enum SeaportOrderKind {
```
georgeroman, line 5: The order kind should belong to the indexer package and not to the sdk.

xcaptain: Ok, will think twice to see how to make offchainCheck support both order kind

PR #3880, opened 2023-03-31
d3or/add tags to websocket events

d3or:
- feat: add tags to websocket events
- feat: tweak token set id splitting

### DIFF
--- a/packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
+++ b/packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
@@ -124,9 +124,17 @@ export class NewSellOrderWebsocketEvent {
       };
 
       redisWebsocketPublisher.publish(
-        "asks",
+        "events",
         JSON.stringify({
           event: "ask.created",
+          tags: {
+            collectionId: rawResult.contract,
+            token_set_id: rawResult.token_set_id,
+            tokenId: rawResult.token_set_id.split(":")[2],
+            kind: rawResult.kind,
+            side: rawResult.side,
+            source: source?.getTitle(),
+          },
           data: result,
         })
       );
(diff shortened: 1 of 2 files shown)


packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
```diff
@@ -124,9 +124,17 @@ export class NewSellOrderWebsocketEvent {
       };
 
       redisWebsocketPublisher.publish(
-        "asks",
+        "events",
```
nofir, line 127: Does users still need to signup to "events" channel or this is just an internal impl?

packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
```diff
@@ -124,9 +124,17 @@ export class NewSellOrderWebsocketEvent {
...
         JSON.stringify({
           event: "ask.created",
+          tags: {
+            collectionId: rawResult.contract,
```
nofir, line 131: the collection id is not always the contract.

packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
```diff
@@ -124,9 +124,17 @@ export class NewSellOrderWebsocketEvent {
...
+            tokenSetId: rawResult.token_set_id,
+            tokenId: rawResult.token_set_id.split(":")[2],
+            kind: rawResult.kind,
+            side: rawResult.side,
```
nofir, line 135: side will always be sell for asks so its redundant

packages/indexer/src/jobs/websocket-events/events/new-sell-order-websocket-event.ts
d3or: this is just for internal because a channel is required to publish to

nofir (approved):

PR #3903, opened 2023-04-03
fix: improve gettokens perf with nativeSource

robertistok:
The modified query includes a subquery in the WHERE clause that filters tst.token_id. This subquery retrieves the list of token_id values from the orders table that meet the specified conditions (sell side, fillable, approved, source_id_int, and taker criteria). The main query then filters tst.token_id based on the results of this subquery.

The reason why this modification might lead to improved performance is that it narrows down the set of token_id values that need to be considered during the join between the orders table and the token_sets_tokens table. By filtering the tst.token_id based on the subquery, the query planner can potentially eliminate a large number of unnecessary rows before executing the join operation.


ptrwtts:
@ipeleg / @nofir / @tv3636 lets get this merged. @robertistok has done some testing to confirm that it seems to improve performance in all cases, without hurting it in others

ipeleg (approved):

PR #3923, opened 2023-04-03
Jaden/fix get api key details permissions return type

JadenDurnford:


packages/indexer/src/api/endpoints/admin/get-api-key-details.ts
```diff
@@ -55,7 +55,7 @@ export const getApiKeyDetails: RouteOptions = {
         email: apiKey.email,
         active: apiKey.active,
         tier: apiKey.tier,
-        permissions: apiKey.permissions,
+        permissions: apiKey.permissions ? JSON.stringify(apiKey.permissions) : null,
```
nofir, line 58: @JadenDurnford ,
- what is the purpose of adding this? are we going to display on the dev portal?
- Why not return it as an object rather than a string?

JadenDurnford: It was causing errors as the return type didn't match (if permissions existed on the key the request failed). I wasn't sure if the object parameters for permissions were concrete across all API keys, but if they are I can return it as an object

JadenDurnford: Yup you're correct I just made this change. I thought I would need to specify the parameters of the object but it is working fine without that.

PR #3955, opened 2023-04-04
Feat/improve sync sales

pedromcunha:


packages/indexer/src/api/endpoints/sync/sync-sales/v1.ts
```diff
@@ -60,6 +73,31 @@ export const getSyncSalesV1Options: RouteOptions = {
...
+        (query as any).contractsFilter.push(contractsFilter);
+      }
+
+      (query as any).contractsFilter = _.join((query as any).contractsFilter, ",");
+      contractFilter = `fill_events_2.contract IN ($/contractsFilter:raw/)`;
+    }
+
+    if (query.monthTimestamp) {
+      (query as any).monthDateStart = format(new Date(query.monthTimestamp), "yyyy-MM-01");
+      (query as any).monthDateEnd = format(lastDayOfMonth(query.monthTimestamp), "yyyy-MM-dd");
```
nofir, line 97: Im not sure but wont this miss sales on the last day of the month? should it updated_at < first day of next month?

packages/indexer/src/api/endpoints/sync/sync-sales/v1.ts
```diff
@@ -95,25 +133,43 @@ export const getSyncSalesV1Options: RouteOptions = {
...
+
+      if (!continuation) {
+        let isCurrentMonth = true;
+        if (query.monthTimestamp) {
+          const monthDateStart = new Date(query.monthTimestamp);
+          const currentMonthDateStart = new Date();
+          isCurrentMonth = isSameMonth(monthDateStart, currentMonthDateStart);
+        }
+
+        cursor = isCurrentMonth ? continuationToken : null;
```
nofir, line 170: Can you explain again the logic behind having both continuation and cursor?

pedromcunha: So essentially the cursor should only be returned for the very last month, to let the light node know where it should put a bookmark to pick back up in the next sync. For months in the past (not the current month), there shouldn't be any changes.

packages/indexer/src/api/endpoints/sync/sync-sales/v1.ts
pedromcunha: It shouldn't because it's just looking at the last day (not including mins/seconds). But to be sure I'll go ahead and add that just in case.

pedromcunha: Should be fixed now.

nofir (approved):

PR #3976, opened 2023-04-05
feat: added option to filter sales endpoint by a list of tokens

arthurmr96:


packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -35,6 +35,12 @@ export const getSalesV4Options: RouteOptions = {
         .description(
           "Filter to a particular token. Example: `0x8d04a8c79ceb0889bdd12acdf3fa9d207ed3ff63:123`"
         ),
+      tokens: Joi.array()
```
nofir, line 38: I would avoid having both token and tokens.
You can either:
- add a new v5 endpoint
- change the existing one to have "token" support both options. similar to here: https://github.com/reservoirprotocol/indexer/blob/578910fb5fd19ecd4685aef45f1a50e6b031c638/packages/indexer/src/api/endpoints/collections/get-collections/v5.ts#L57
-

packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -123,6 +130,25 @@ export const getSalesV4Options: RouteOptions = {
       (query as any).contract = toBuffer(contract);
       (query as any).tokenId = tokenId;
       tokenFilter = `fill_events_2.contract = $/contract/ AND fill_events_2.token_id = $/tokenId/`;
+    } else if (query.tokens) {
```
nofir, line 133: See comment above, you can merge both token and tokens logic

nofir (changes requested):
left some comments

packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
arthurmr96: I added a new version then, so the property can be called `tokens` instead of `token`. It makes makes more sense to me the name be this one to support both options

nofir (approved):
LGTM. We will do some testing and merge.

nofir (changes requested):

packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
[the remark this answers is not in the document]
arthurmr96: done!

ipeleg:
@arthurmr96 we're missing 
  server.route({
    method: "GET",
    path: "/sales/v5",
    options: transfersEndpoints.getSalesV5Options,
  });

in the routes.ts, can you add it?

PR #3984, opened 2023-04-06
feat: support Looksrare v2

lljxx1:


packages/sdk/src/looks-rare-v2/builders/contract-wide/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    } catch {
+      return false;
+    }
+
+    return true;
+  }
+
+  public build(params: BuildParams) {
+    // if (params.isOrderAsk) {
```
georgeroman, line 31: We can remove the commented code.

packages/sdk/src/looks-rare-v2/builders/contract-wide/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      strategyId: 1,
+      currency: params.currency,
+      quoteType: params.quoteType,
+      collectionType: params.collectionType,
+
+      startTime: params.startTime!,
+      endTime: params.endTime!,
+      additionalParameters: params.additionalParameters ?? BytesEmpty,
+
+      globalNonce: params.globalNonce ? s(params.globalNonce) : "0",
```
georgeroman, line 53: Let's move the initialization of the nonces in the above `defaultInitialize` method.

packages/sdk/src/looks-rare-v2/builders/base/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  public chainId: number;
+
+  constructor(chainId: number) {
+    this.chainId = chainId;
+  }
+
+  protected defaultInitialize(params: BaseBuildParams) {
+    params.startTime = params.startTime ?? getCurrentTimestamp(-1 * 60);
+    params.endTime = params.endTime ?? getCurrentTimestamp(24 * 60 * 60);
+    // params.minPercentageToAsk = params.minPercentageToAsk ?? 8500;
```
georgeroman, line 42: Let's remove the commented code.

georgeroman (changes requested):

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,268 @@
...
+
+        break;
+      }
+
+      case "looks-rare-v2-taker-bid": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const orderId = parsedLog.args["nonceInvalidationParameters"]["orderHash"].toLowerCase();
+        const orderNonce = parsedLog.args["nonceInvalidationParameters"]["orderNonce"].toString();
+        const maker = parsedLog.args["bidUser"].toLowerCase();
+        let taker = parsedLog.args["bidRecipient"].toLowerCase();
```
georgeroman, line 180: Should this be `askUser` instead?

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,268 @@
...
+        break;
+      }
+
+      case "looks-rare-v2-taker-ask": {
+        const parsedLog = eventData.abi.parseLog(log);
+
+        const orderId = parsedLog.args["nonceInvalidationParameters"]["orderHash"].toLowerCase();
+        const orderNonce = parsedLog.args["nonceInvalidationParameters"]["orderNonce"].toString();
+
+        const maker = parsedLog.args["askUser"].toLowerCase();
```
georgeroman, line 84: The maker and taker are swapped. This is a `taker-ask`, so the maker is the bid user.

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        break;
+      }
+
+      case "looks-rare-v2-taker-bid": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const orderId = parsedLog.args["nonceInvalidationParameters"]["orderHash"].toLowerCase();
+        const orderNonce = parsedLog.args["nonceInvalidationParameters"]["orderNonce"].toString();
+        const maker = parsedLog.args["bidUser"].toLowerCase();
```
georgeroman, line 179: The maker and taker are swapped. This is a `taker-bid`, so the maker is the ask user.

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,268 @@
...
+          orderSide: "buy",
+          contract,
+          tokenId,
+          amount,
+          price: priceData.nativePrice,
+          timestamp: baseEventParams.timestamp,
+          maker,
+          taker,
+        });
+
```
georgeroman, line 171: Since this is a bid acceptance sale, we should add in the following piece of code which handles erc20 changes:
https://github.com/reservoirprotocol/indexer/blob/de83e3add90347651bd1b9c032389f80f2e531db/packages/indexer/src/sync/events/handlers/zeroex-v4.ts#L336-L354

packages/indexer/src/orderbook/orders/looks-rare-v2/check.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // Check: order is not filled
+    const quantityFilled = await commonHelpers.getQuantityFilled(id);
+    if (quantityFilled.gte(order.params.amounts[0])) {
+      throw new Error("filled");
+    }
+  }
+
+  // Check: order's nonce was not bulk cancelled
```
georgeroman, line 45: We should uncomment this, and add support for checking the min nonce by side (which LR uses).

georgeroman (changes requested):

packages/contracts/test/sdk/looks-rare-v2/contract-wide/erc1155.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      collectionType: LooksRareV2.Types.CollectionType.ERC1155,
+      signer: buyer.address,
+      collection: erc1155.address,
+      itemId: 0,
+      amount: 1,
+      currency: Common.Addresses.Weth[chainId],
+      price,
+      startTime: await getCurrentTimestamp(ethers.provider),
+      endTime: (await getCurrentTimestamp(ethers.provider)) + 60,
+      orderNonce: await exchange.getNonce(ethers.provider, buyer.address, "buy"),
```
georgeroman, line 82: As far as I understand, the order nonce can be anything (any random number that wasn't used before). This is different from the global nonce, which is what seems to be retrieved via the `getNonce` method. So in order to not mislead people reading these tests, we should simply set the order nonce to a random value here. Same applies to all tests.

packages/indexer/src/orderbook/orders/common/helpers.ts
```diff
@@ -103,22 +103,29 @@ export const getNftApproval = async (
   return approvalResult ? approvalResult.approved : false;
 };
 
-export const getMinNonce = async (orderKind: OrderKind, maker: string): Promise<BigNumber> => {
+export const getMinNonce = async (
+  orderKind: OrderKind,
+  maker: string,
+  side?: string
```
georgeroman, line 109: The `side` field should have a type of `"sell" | "buy"` instead of `string`.

packages/indexer/src/orderbook/orders/common/helpers.ts
```diff
@@ -103,22 +103,29 @@ export const getNftApproval = async (
...
+): Promise<BigNumber> => {
   const bulkCancelResult: { nonce: string } | null = await idb.oneOrNone(
     `
       SELECT coalesce(
         (
           SELECT bulk_cancel_events.min_nonce FROM bulk_cancel_events
           WHERE bulk_cancel_events.order_kind = $/orderKind/
             AND bulk_cancel_events.maker = $/maker/
+            AND bulk_cancel_events.side IS NULL OR bulk_cancel_events.side = $/side/
```
georgeroman, line 118: Since `side` can be `undefined` (and this code will always result in an SQL statement), let's change this to `side ? " AND bulk_cancel_events.side = $/side/" : ""`.

packages/indexer/src/orderbook/orders/looks-rare-v2/build/buy/collection.ts
```diff
@@ -0,0 +1,36 @@
+import * as Sdk from "@reservoir0x/sdk";
+import { BaseBuilder } from "@reservoir0x/sdk/dist/looks-rare/builders/base";
```
georgeroman, line 2: This should import from the `looks-rare-v2` SDK package. Same applies to all below build methods.

packages/indexer/src/orderbook/orders/looks-rare-v2/check.ts
```diff
@@ -0,0 +1,130 @@
...
+  const side = order.params.quoteType === Sdk.LooksRareV2.Types.QuoteType.Ask ? "sell" : "buy";
+  const minNonce = await commonHelpers.getMinNonce("looks-rare-v2", order.params.signer, side);
+  if (minNonce.gt(order.params.orderNonce)) {
```
georgeroman, line 48: We should compare against the order's `globalNonce`, right?

packages/indexer/src/sync/events/data/looks-rare-v2.ts
```diff
@@ -0,0 +1,268 @@
...
+  numTopics: 1,
+  abi: new Interface([
+    `event OrderNoncesCancelled(
+      address user, 
+      uint256[] orderNonces
+    )`,
+  ]),
+};
+
+export const subsetNoncesCancelled: EventData = {
```
georgeroman, line 36: I need to do a bit more research into what these subset nonces represent. We might have to build something custom to properly handle it.

packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -0,0 +1,425 @@
...
+        onChainRoyalties = await royalties.getRoyalties(
+          order.params.collection,
+          order.params.itemIds[0],
+          "onchain"
+        );
+      } else {
+        onChainRoyalties = await royalties.getRoyaltiesByTokenSet(tokenSetId, "onchain");
+      }
+
+      // TODO: Remove (for backwards-compatibility only)
```
georgeroman, line 209: I think we can remove this, as the comment suggests.

packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -0,0 +1,425 @@
...
+          ...feeBreakdown,
+          {
+            kind: "royalty",
+            recipient: onChainRoyalties[0].recipient,
+            // LooksRare has fixed 0.5% royalties
+            bps: 50,
+          },
+        ];
+      } else {
+        // If there is no royalty, the marketplace fee will be 2%
```
georgeroman, line 233: Do the same rules as with v1 apply to v2?

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,317 @@
...
+  for (const { subKind, baseEventParams, log } of events) {
+    if (currentTx !== baseEventParams.txHash) {
+      currentTx = baseEventParams.txHash;
+      currentTxLogs = [];
+    }
+    currentTxLogs.push(log);
+
+    const eventData = getEventData([subKind])[0];
+    switch (subKind) {
+      case "looks-rare-v2-new-bid-ask-nonces": {
```
georgeroman, line 25: The bid/ask nonces, aka global nonces, should be handled like we handle it for `element`, with `acrossAll` set to true:
https://github.com/reservoirprotocol/indexer/blob/42cf8d6b42d270efc805f9a412456e143afa87f7/packages/indexer/src/sync/events/handlers/element.ts#L52

packages/indexer/src/sync/events/data/looks-rare-v2.ts
georgeroman: image

georgeroman: I think we can easily handle the global nonce and the order nonce, but we need to add support for subset nonces. Instead of adding a new column in the `orders` table I think what we can do when a subset nonce gets cancelled is:
- iterate through all the maker's LR v2 orders which are fillable (the query should use the below index)
- cancel all orders which have in the raw data the currently cancelled subset nonce

image

In addition, we should create a new table `looksrare_v2_subset_nonces` which keeps track of the cancelled subset nonces per user and make sure to check it in the `offChainCheck` method so that we don't ingest any already-cancelled orders.

georgeroman (changes requested):

georgeroman:
One last comment, LR v2 also supports bulk signatures, like Seaport v1.4. We should support at least verifying bulk-signed orders.

packages/indexer/src/sync/events/data/looks-rare-v2.ts
lljxx1: `orderNonce` cancel and with the `subsetNonce` cancel will be in a different tx,  so I reuse the `nonce_cancel_events`. add `isSubset` to once cancel event and generate different sql

``` sql
${
        isSubset
          ? `
          AND ("o"."raw_data"->>'subsetNonce')::NUMERIC = "x"."nonce"
        `
          : `  AND "o"."nonce" = "x"."nonce" `
      }
```

packages/indexer/src/orderbook/orders/common/helpers.ts
```diff
@@ -103,22 +103,31 @@ export const getNftApproval = async (
...
+  side?: "sell" | "buy"
+): Promise<BigNumber> => {
   const bulkCancelResult: { nonce: string } | null = await idb.oneOrNone(
     `
       SELECT coalesce(
         (
           SELECT bulk_cancel_events.min_nonce FROM bulk_cancel_events
           WHERE bulk_cancel_events.order_kind = $/orderKind/
             AND bulk_cancel_events.maker = $/maker/
+            ` +
```
georgeroman, line 118: Easier to read this way: `${side ? " AND bulk_cancel_events.side = $/side/" : ""}`

packages/sdk/src/looks-rare-v2/builders/base/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  collectionType: CollectionType;
+
+  collection: string;
+  currency: string;
+  price: BigNumberish;
+  signer: string;
+
+  startTime: number;
+  endTime: number;
+  itemId: BigNumberish;
```
georgeroman, line 23: I don't think `itemId` should be part of the base build params since it's only relevant for `single-token` orders and not `contract-wide`. Also, let's remove the amount since I think LR v2 current strategies only support an amount of 1 so we can just hardcode that for now (until they add support for amounts greater than 1).

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,341 @@
...
+
+        break;
+      }
+
+      case "looks-rare-v2-taker-bid": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const orderId = parsedLog.args["nonceInvalidationParameters"]["orderHash"].toLowerCase();
+        const orderNonce = parsedLog.args["nonceInvalidationParameters"]["orderNonce"].toString();
+        const maker = parsedLog.args["bidRecipient"].toLowerCase();
+        let taker = parsedLog.args["bidUser"].toLowerCase();
```
georgeroman, line 229: I think `taker` should be `bidRecipient` while `maker` should be `askUser`.

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,341 @@
...
+          contract,
+          tokenId,
+          amount,
+          price: priceData.nativePrice,
+          timestamp: baseEventParams.timestamp,
+          maker,
+          taker,
+        });
+
+        // then we need resync the maker's ERC20 approval to the exchange
```
georgeroman, line 318: This was only needed for `taker-ask`.

georgeroman (changes requested):

packages/indexer/src/jobs/arweave-relay/index.ts
```diff
@@ -0,0 +1,286 @@
+import * as Sdk from "@reservoir0x/sdk";
```
georgeroman, line 1: Can you rebase on top of main? This file doesn't exist anymore.

packages/sdk/src/looks-rare-v2/builders/single-token/index.ts
```diff
@@ -0,0 +1,68 @@
+import { BaseBuildParams, BaseBuilder } from "../base";
+import { Order } from "../../order";
+import { BytesEmpty, s } from "../../../utils";
+import { BigNumberish } from "@ethersproject/bignumber";
+
+interface BuildParams extends BaseBuildParams {
+  itemId: BigNumberish;
+  amount: BigNumberish;
```
georgeroman, line 8: For now I think we don't need `amount` but can just hardcoded it to be 1.

packages/indexer/src/sync/events/storage/nonce-cancel-events.ts
```diff
@@ -25,10 +26,67 @@ type DbEvent = {
   nonce: string;
 };
 
+function generateUpdateQuery(nonceCancelValues: DbEvent[], isSubset: boolean) {
```
georgeroman, line 29: As I mentioned in Slack, I don't think this will work. We probably need something custom:
image

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
georgeroman: @lljxx1 I don't think this is resolved.

packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -0,0 +1,424 @@
...
+      if (listingTime - 5 * 60 >= currentTime) {
+        // TODO: Add support for not-yet-valid orders
+        return results.push({
+          id,
+          status: "invalid-listing-time",
+        });
+      }
+
+      // Mutiple tokens
+      if (order.params.itemIds.length > 1) {
```
georgeroman, line 73: We should also skip any orders which have an amount greater than 1 (at least for now).

packages/sdk/src/looks-rare-v2/types.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  startTime: number;
+  endTime: number;
+  price: string;
+  itemIds: string[];
+  amounts: string[];
+
+  additionalParameters: string;
+  signature?: string;
+  merkletree?: MerkleTree;
```
georgeroman, line 62: Let's use camel case: `merkleTree`

packages/indexer/src/migrations/1680570377134_subset-nonces.sql
```diff
@@ -0,0 +1,29 @@
+-- Up Migration
+
+CREATE TABLE "subset_nonce_events" (
```
georgeroman, line 3: Since this is a LR v2 specific logic, let's rename it (and the filename + associated indexes) to `looksrare_v2_subset_nonce_cancel_events`.

packages/indexer/src/migrations/1680570377134_subset-nonces.sql
```diff
@@ -0,0 +1,29 @@
...
+  "address" BYTEA NOT NULL,
+  "block" INT NOT NULL,
+  "block_hash" BYTEA NOT NULL,
+  "tx_hash" BYTEA NOT NULL,
+  "tx_index" INT NOT NULL,
+  "log_index" INT NOT NULL,
+  "timestamp" INT NOT NULL,
+  "batch_index" INT NOT NULL,
+  "order_kind" order_kind_t NOT NULL,
```
georgeroman, line 12: Since this table is LR v2 specific, we don't need the order kind here.

packages/indexer/src/migrations/1680570377134_subset-nonces.sql
```diff
@@ -0,0 +1,29 @@
...
+  "maker" BYTEA NOT NULL,
+  "nonce" NUMERIC(78, 0) NOT NULL
+);
+
+ALTER TABLE "subset_nonce_events"
+  ADD CONSTRAINT "subset_nonce_events_pk"
+  PRIMARY KEY ("tx_hash", "log_index", "batch_index");
+
+CREATE INDEX "subset_nonce_events_block_block_hash_index"
+  ON "subset_nonce_events" ("block", "block_hash");
```
georgeroman, line 22: No need for this index.

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -0,0 +1,320 @@
...
+      }
+
+      case "looks-rare-v2-subset-nonces-cancelled": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const maker = parsedLog.args["user"].toLowerCase();
+        const subsetNonces = parsedLog.args["subsetNonces"].map(String);
+
+        let batchIndex = 1;
+        for (const subsetNonce of subsetNonces) {
+          onChainData.subsetNonceCancelEvents.push({
```
georgeroman, line 66: Since this logic is LR v2 specific, I think we can just have all logic (eg. packages/indexer/src/sync/events/storage/subset-nonce-cancel-events.ts) directly integrated here:

packages/indexer/src/sync/events/storage/nonce-cancel-events.ts
```diff
@@ -25,10 +25,60 @@ type DbEvent = {
   nonce: string;
 };
 
+function generateUpdateQuery(nonceCancelValues: DbEvent[]) {
```
georgeroman, line 28: Let's revert this change since nothing actually changed.

georgeroman (changes requested):

packages/indexer/src/sync/events/handlers/looks-rare-v2.ts
```diff
@@ -62,18 +78,117 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+              "log_index",
+              "timestamp",
+              "batch_index",
+              "maker",
+              "nonce"
+            ) VALUES ${pgp.helpers.values(nonceCancelValues, columns)}
+            ON CONFLICT DO NOTHING
+            RETURNING "maker", "nonce", "tx_hash", "timestamp", "log_index", "batch_index", "block_hash"
+          )
+          SELECT
```
georgeroman, line 134: For full atomicity we should do the order update atomically, something like below:
```sql
WITH
x AS (
   ...
),
UPDATE orders SET
            "fillability_status" = 'cancelled',
            "expiration" = to_timestamp("x"."timestamp"),
            "updated_at" = now()
          FROM x
          WHERE "o"."kind" = 'looks-rare-v2'
            AND "o"."maker" = "x"."maker"
            AND ("o"."fillability_status" = 'fillable' OR "o"."fillability_status" = 'no-balance')
            AND ("o".raw_data"->>'subsetNonce'::NUMERIC(78,0) IN $/nonceCancelValues/
```

georgeroman (changes requested):

georgeroman:
@lljxx1 Ready to go to dev for testing!

lljxx1:
@georgeroman based changed to dev

PR #3998, opened 2023-04-06
feat: adding index for updated_at asc on active listings

JadenDurnford:


packages/indexer/src/migrations/1643809164129_orders.sql
```diff
@@ -135,6 +135,10 @@ CREATE INDEX "orders_asks_updated_at_asc_id_index"
   ON "orders" ("updated_at" ASC, "id" ASC)
   WHERE ("side" = 'sell');
 
+-- CREATE INDEX "orders_updated_at_asc_id_active_index"
+--   ON "orders" ("side", "updated_at" ASC, "id" ASC)
```
nofir, line 139: the index was created, you can uncomment this

nofir (changes requested):

PR #4050, opened 2023-04-11
fix: removed infinity marketplace

jfrazier-eth:
Infinity has been deprecated

I left the Infinity sdk implementation since that is a public npm package and is still fully functional.


packages/indexer/src/api/endpoints/orders/post-orders/v1.ts
```diff
@@ -38,9 +38,7 @@ export const postOrdersV1Options: RouteOptions = {
               "blur",
               "rarible",
               "manifold",
-              "infinity",
-              "flow",
-              "looks-rare-v2"
```
georgeroman, line 43: We still want `looks-rare-v2`

georgeroman (changes requested):
Can you also remove the relevant code in the `relayer` package?

jfrazier-eth:
> Can you also remove the relevant code in the `relayer` package?

https://github.com/reservoirprotocol/relayer/pull/81
Is this still the correct repo?

PR #4070, opened 2023-04-12
feat: add cacheCollectionTopBidValue logic

d3or:


packages/indexer/src/orderbook/orders/seaport-v1.4/index.ts
```diff
@@ -622,22 +622,52 @@ export const save = async (
...
+              // if the bid is higher than the top bid cache it
+
+              const expiry = new Date();
+
+              // set expiry as seconds until order.params.endTime
+              expiry.setSeconds(order.params.endTime - currentTime);
+
+              const seconds = expiry.getSeconds();
+
+              await cacheCollectionTopBidValue(
```
nofir, line 649: I am not sure this is correct, the top bid on a single token is not necessarily the collection top bid. The caching should maybe be here: https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/jobs/collection-updates/top-bid-queue.ts

nofir (changes requested):

packages/indexer/src/jobs/collection-updates/top-bid-queue.ts
```diff
@@ -130,6 +135,22 @@ if (config.doBackgroundWork) {
           }
         );
 
+        // cache the new top bid
+
+        const expiry = new Date();
```
nofir, line 140: I think you're missing a `if (collectionTopBid?.order_id)`, and then and `else` with removing the current collection top bid from the cache.

packages/indexer/src/orderbook/orders/seaport-v1.4/index.ts
```diff
@@ -622,23 +622,37 @@ export const save = async (
       if (info.side === "buy" && order.params.kind === "single-token" && validateBidValue) {
         const typedInfo = info as typeof info & { tokenId: string };
         const tokenId = typedInfo.tokenId;
-        const seaportBidPercentageThreshold = 80;
+        const seaportBidPercentageThreshold = 50;
```
nofir, line 625: lets keep it 80 for the initial deployment and then make the change, i worry its going to overwhelm the system

nofir (changes requested):

packages/indexer/src/jobs/collection-updates/top-bid-queue.ts
```diff
@@ -130,6 +138,28 @@ if (config.doBackgroundWork) {
...
 
+        if (collectionTopBid?.order_id) {
+          // cache the new top bid
+
+          const expiry = new Date();
+          // set redis expiry as seconds until the top bid expires
+          expiry.setSeconds(collectionTopBid?.valid_until - now());
+          const seconds = expiry.getSeconds();
+
+          const [, , tokenId] = collectionTopBid.token_set_id.split(":");
```
nofir, line 149: i dont think the token set will have the token id in it, given that this logic only handles collection offers.

packages/indexer/src/jobs/collection-updates/top-bid-queue.ts
```diff
@@ -106,6 +110,10 @@ if (config.doBackgroundWork) {
               y.top_buy_id,
               y.top_buy_source_id_int,
               y.top_buy_valid_between,
+              coalesce(
```
nofir, line 113: Are you sure its going to work? this is not included in the RETURNING clause (i am

packages/indexer/src/jobs/collection-updates/top-bid-queue.ts
nofir: the collection id is actually passed to the job, why not just use that?

nofir (changes requested):

nofir (dismissed):

ipeleg (dismissed):

ipeleg (approved):

PR #4121, opened 2023-04-14
feat: add orderKind when calling oracle

xcaptain:
See also: https://github.com/reservoirprotocol/seaport-oracle/pull/3

Close #4116


packages/indexer/src/orderbook/orders/seaport-v1.4/index.ts
```diff
@@ -686,6 +686,7 @@ export const save = async (
             {
               newOrders: [order.params],
               replacedOrders: [replacedOrderResult.raw_data],
+              orderKind: "alienswap",
```
georgeroman, line 689: This should be 'seaport-v1.4'

xcaptain: Thanks, fixed the typo

PR #4123, opened 2023-04-14
feat: add sell and buy order queue

d3or:


packages/indexer/src/jobs/order-updates/buy-order-queue.ts
```diff
@@ -0,0 +1,412 @@
...
+                txHash: trigger.txHash || null,
+                txTimestamp: trigger.txTimestamp || null,
+              })
+            );
+          }
+        }
+
+        if (buyOrderResult.length) {
+          if (
+            trigger.kind === "new-order" &&
```
nofir, line 163: we need to change it to enable it for single token bid

nofir (changes requested):

ipeleg (approved):

packages/indexer/src/jobs/order-updates/by-id-queue.ts
```diff
@@ -47,7 +38,7 @@ if (config.doBackgroundWork) {
   worker = new Worker(
     QUEUE_NAME,
     async (job: Job) => {
-      const { id, trigger } = job.data as OrderInfo;
```
tv3636, line 50: can you merge main/add this back in? it's not going to catch it as a merge conflict but the log i added for tracking order ingestion uses trigger in this job now

ipeleg (approved):

PR #4185, opened 2023-04-19
feat: extract sub-call attribution data and fix royalytie issue

lljxx1:


georgeroman:
@lljxx1 The royalty part looks good but regarding the calldata attribution I think we want to it inline in this method rather than having a separate process that does it asynchonously:
https://github.com/reservoirprotocol/indexer/blob/4ba1ef38d0089407391aa710a562015d62444102/packages/indexer/src/sync/events/utils.ts#L156

packages/indexer/src/sync/events/utils.ts
```diff
@@ -153,8 +154,12 @@ export const fetchTransactionLogs = async (txHash: string) =>
     });
   });
 
-export const extractAttributionData = async (
-  txHash: string,
+export const extractAttributionDataByTransaction = async (
```
georgeroman, line 157: I'd like to keep this attribution method as it was before, since otherwise it will require more thorough testing. What I think we need is just including the below logic after line 188:
image

packages/indexer/src/jobs/fill-updates/fill-post-process.ts
```diff
@@ -52,6 +52,10 @@ if (config.doBackgroundWork) {
                 marketplace_fee_breakdown = $/marketplaceFeeBreakdown:json/,
                 paid_full_royalty = $/paidFullRoyalty/,
                 net_amount = $/netAmount/,
+                order_source_id_int = $/orderSourceId/,
```
georgeroman, line 55: Let's remove these, since for now we're going to set these fields synchronously

packages/indexer/src/sync/events/handlers/royalties/index.ts
```diff
@@ -89,6 +108,29 @@ export const getFillEventsFromTx = async (txHash: string): Promise<PartialFillEv
   );
 };
 
+export const getOrderInfos = async (orderIds: string[]): Promise<OrderInfo[]> => {
+  if (!orderIds.length) return [];
+  const results = await idb.manyOrNone(
```
georgeroman, line 113: Can we use `redb` here?

georgeroman (changes requested):

PR #4211, opened 2023-04-20
d3or/stop single token bids

d3or:
- feat: add sell and buy order queue
- fix: move websocket event router into their own seperate jobs
- feat: rename queues and remove redundant order querying
- fix: fix fields to be converted from bytea to buffer in queue transport
- fix: fix job data logging
- fix: fix job data logging
- feat: throw error when order not found
- fix: fix order to buffer to be inside if order condition
- fix: remove newtokenbid eventkind
- feat: add condition to stop single token bids in case of perf issues


ipeleg:
Closing this for now, as this one is not relevant any more

PR #4212, opened 2023-04-21
feat: check if granted when listing/biding

lljxx1:


georgeroman:
@lljxx1 Just a couple of comments:
- can you add this to the v4 and v3 APIs as well (those also support LR v2)
- can you change the branch to development (we should test on goerli first)

PR #4246, opened 2023-04-24
add min and max floor price filters to the collections endpoint

lunarmayor:
added min and max ask price filters for the collections endpoint for an experiment and think it would be useful for everyone


nofir (approved):
Looks good, we'll test and deploy

PR #4270, opened 2023-04-25
Adding GXB polygon currency

Nike682631:
https://docs.reservoir.tools/docs/adding-support-for-custom-erc-20-token

### DIFF
--- a/packages/indexer/src/config/network.ts
+++ b/packages/indexer/src/config/network.ts
@@ -385,6 +385,15 @@ export const getNetworkSettings = (): NetworkSettings => {
               decimals: 18,
             },
           ],
+          [
+            "0x3b45a986621f91eb51be84547fbd9c26d0d46d02",
+            {
+              contract: "0x3B45a986621f91Eb51bE84547fBd9c26D0D46D02",
+              name: "Gold Bar Currency",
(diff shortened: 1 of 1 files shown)


packages/indexer/src/config/network.ts
```diff
@@ -385,6 +385,15 @@ export const getNetworkSettings = (): NetworkSettings => {
               decimals: 18,
             },
           ],
+          [
+            "0x3b45a986621f91eb51be84547fbd9c26d0d46d02",
+            {
+              contract: "0x3B45a986621f91Eb51bE84547fBd9c26D0D46D02",
```
ipeleg, line 391: @Nike682631 looks good, one thing, can you change the contract to be lower case?

ipeleg (changes requested):

packages/indexer/src/config/network.ts
Nike682631: @ipeleg  done! btw you should consider mentioning in this [link](https://docs.reservoir.tools/docs/adding-support-for-custom-erc-20-token) that both contract addresses need be to lowercase.

ipeleg (approved):

PR #4274, opened 2023-04-26
fix: royalties detection support multi-curreny

lljxx1:


georgeroman:
Some entries might not have `currencyPrice` set, and in this case we want to default to `price` so instead of `currencyPrice!` we should use `currencyPrice ?? price`.

PR #4297, opened 2023-04-27
feat: support opensea 0x on polygon

lljxx1:


packages/indexer/src/sync/events/data/opensea-zeroex-v3.ts
```diff
@@ -0,0 +1,31 @@
+import { Interface } from "@ethersproject/abi";
+import { OpenSeaZeroExV3 } from "@reservoir0x/sdk";
+
+import { config } from "@/config/index";
+import { EventData } from "@/events-sync/data";
+
+export const fill: EventData = {
+  kind: "opensea-zeroex-v3",
```
georgeroman, line 8: We already have support for handling this event (in the `zeroex-v3` handler which is here https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/sync/events/data/zeroex-v3.ts). It seems that the address is wrong (https://github.com/reservoirprotocol/indexer/blob/main/packages/sdk/src/zeroex-v3/addresses.ts#L4), but we should simply update that to `0xfede379e48c873c75f3cc0c81f7c784ad730a8f7` and then have the logic in the existing `zeroex-v3` handler.

georgeroman (changes requested):

PR #4306, opened 2023-04-28
feat: update related timestamp

lljxx1:


packages/indexer/src/jobs/backfill/backfill-block-timestamps.ts
```diff
@@ -49,13 +49,46 @@ if (config.doBackgroundWork) {
       const columns = new pgp.helpers.ColumnSet(["number", "timestamp"], {
         table: "blocks",
       });
+
+      const blockAndTimestamps: any[] = [];
```
georgeroman, line 53: We know the type here, let's avoid `any` where it's not that inconvenient to do that.

packages/indexer/src/jobs/backfill/backfill-block-timestamps.ts
```diff
@@ -49,13 +49,46 @@ if (config.doBackgroundWork) {
...
+        } else {
+          blockAndTimestamps.push({
+            number,
+            timestamp,
+          });
         }
       }
 
+      // Update related wrong timestamp data
+      await Promise.all([
```
georgeroman, line 71: For both `nft_transfer_events` and `fill_events_2` let's also set the `updated_at` column.

georgeroman (changes requested):

PR #4340, opened 2023-05-01
Add support for $PEPE

dturska12:


nofir:
@dturska12 We use coingecko for supported currencies. Looks like coingecko support this currency (https://www.coingecko.com/en/coins/pepe) so there is no need to add custom support for it.

nofir:
Closing as this currency is already supported

PR #4366, opened 2023-05-03
feat: add buy/sale websocket events

d3or:


packages/indexer/src/jobs/websocket-events/buy-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,175 @@
...
+          txHash: r.tx_hash,
+          logIndex: r.log_index,
+          batchIndex: r.batch_index,
+          isDeleted: Boolean(r.is_deleted),
+          createdAt: new Date(r.created_at).toISOString(),
+          updatedAt: new Date(r.updated_ts * 1000).toISOString(),
+        });
+
+        let eventType = "";
+        if (data.trigger === "insert") eventType = "buy.created";
```
nofir, line 124: What is buy? Why dont just have a sale websocket? we have the side and the taker and the maker.

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,175 @@
...
+          orderKind: r.order_kind,
+          maker: r.maker,
+          taker: r.taker,
+          amount: r.amount,
+          fillSourceId: r.fill_source_id,
+          block: r.block,
+          txHash: r.tx_hash,
+          logIndex: r.log_index,
+          batchIndex: r.batch_index,
+          isDeleted: Boolean(r.is_deleted),
```
nofir, line 118: Should we created a sale.deleted / sale.removed based on is_deleted? wdyt?

d3or: in what instances are is_deleted set to true?

nofir: after we mark it as deleted (usually when we detect an issue or when we revert sales due to orphan blocks)

nofir (approved):
approved but lets see if there is feedback regarding the payload

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,179 @@
...
```
nofir, line 118: you should remove this from the payload.

nofir (changes requested):

PR #4371, opened 2023-05-03
feat: add balance websocket events

d3or:


packages/indexer/src/jobs/websocket-events/balance-event-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+          contract: eventData.contract,
+          tokenId: eventData.token_id,
+          owner: eventData.owner,
+          amount: eventData.amount,
+          acquiredAt: eventData.acquired_at,
+          floorSell: {
+            id: eventData.floor_sell_id,
+            value: eventData.floor_sell_value ? formatEth(eventData.floor_sell_value) : null,
+          },
+          topBid: eventData.top_buy_id
```
nofir, line 50: I dont think we are actually updated this value so might be better to leave out

packages/indexer/src/jobs/websocket-events/balance-event-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+            ? {
+                id: eventData.top_buy_id,
+                value: eventData.top_buy_value ? formatEth(eventData.top_buy_value) : null,
+                maker: eventData.top_buy_maker ? eventData.top_buy_maker : null,
+              }
+            : null,
+        };
+
+        let eventType = "";
+        if (data.trigger === "insert") eventType = "balance.created";
```
nofir, line 60: Im actually not sure about this, maybe it should just be nft-balance. changed for both cases. (except if someone wants to know if the person has it for the first time... )

packages/indexer/src/jobs/websocket-events/balance-event-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+      const { data } = job.data as EventInfo;
+
+      try {
+        const { eventData } = data;
+        const result = {
+          contract: eventData.contract,
+          tokenId: eventData.token_id,
+          owner: eventData.owner,
+          amount: eventData.amount,
+          acquiredAt: eventData.acquired_at,
```
nofir, line 45: im wondering it this makes sense, because for changed it could be that this is not relevant anymore

packages/indexer/src/jobs/websocket-events/balance-event-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+  const worker = new Worker(
+    QUEUE_NAME,
+    async (job: Job) => {
+      const { data } = job.data as EventInfo;
+
+      try {
+        const { eventData } = data;
+        const result = {
+          contract: eventData.contract,
+          tokenId: eventData.token_id,
```
nofir, line 42: make it a token object, so we can add more token info https://github.com/reservoirprotocol/indexer/blob/dc02823a427147bcd09d0dac0adf3c10600085ae/packages/indexer/src/api/endpoints/tokens/get-user-tokens/v7.ts#L630

nofir (changes requested):

nofir (dismissed):

PR #4385, opened 2023-05-03
feat: add transfers websocket event

d3or:


packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,118 @@
...
+if (config.doBackgroundWork && config.doWebsocketServerWork) {
+  const worker = new Worker(
+    QUEUE_NAME,
+    async (job: Job) => {
+      const { data } = job.data as EventInfo;
+
+      try {
+        const { eventData } = data;
+        const result = {
+          address: eventData.address,
```
nofir, line 39: can you match the format of the api (txHash, token obj, etc) see here: 
https://github.com/reservoirprotocol/indexer/blob/05d59ec8029ddbb059d4a0e0b6a01c44bba6d697/packages/indexer/src/api/endpoints/transfers/get-transfers/v3.ts#L245

I see in the api we join with the fill events, need to confirm if that is needed as we will also have.

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,118 @@
...
+          block: eventData.block,
+          txHash: eventData.tx_hash,
+          txIndex: eventData.tx_index,
+          logIndex: eventData.log_index,
+          timestamp: new Date(eventData.timestamp).toISOString(),
+          from: eventData.from,
+          to: eventData.to,
+          tokenId: eventData.token_id,
```
nofir, line 47: see the format here:
https://github.com/reservoirprotocol/indexer/blob/2916a3f7a5536db9eba9ac75cab00498277a5e09/packages/indexer/src/api/endpoints/transfers/get-transfers/v3.ts#L244

i dont think we need the collection info

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
d3or: do you think it might be beneficial to omit things like price / token name / token image? I feel as a large portion of the time these fields will be null because most NFTs dont have liquidity and images (most transfers are when they are minted initially)

nofir: I think it ok and we can add it if needed based on the current structure

nofir (approved):

PR #4409, opened 2023-05-04
feat: add currency symbol and decimals to path for sell and buy

pedromcunha:


georgeroman:
Does it make sense to return the full currency object?

PR #4413, opened 2023-05-05
CollectionXYZ integration

potato-cat69:
Integration of collectionswap protocol with reservoir protocol. Currently does not support bid orders with `externalFilter` set.

### DIFF
--- a/packages/indexer/src/api/endpoints/search/get-search-activities/v1.ts
+++ b/packages/indexer/src/api/endpoints/search/get-search-activities/v1.ts
@@ -159,6 +159,8 @@ export const getSearchActivitiesV1Options: RouteOptions = {
 
     const esQuery = {};
 
+    (esQuery as any).track_total_hits = false;
+
     if (query.types && !_.isArray(query.types)) {
       query.types = [query.types];
     }
@@ -265,12 +267,12 @@ export const getSearchActivitiesV1Options: RouteOptions = {
       (esQuery as any).bool.filter.push(usersFilter);
     }
 
-    let esSort;
+    const esSort: any[] = ["_doc"];
 
     if (query.sortBy == "eventTimestamp") {
-      esSort = [{ timestamp: { order: "desc", missing: "_last" } }];
+      esSort.push({ timestamp: { order: "desc" } });
     } else {
-      esSort = [{ createdAt: { order: "desc" } }];
+      esSort.push({ createdAt: { order: "desc" } });
     }
 
     let searchAfter;
--- a/packages/indexer/src/elasticsearch/indexes/activities/index.ts
+++ b/packages/indexer/src/elasticsearch/indexes/activities/index.ts
@@ -8,8 +8,10 @@ import { SortResults } from "@elastic/elasticsearch/lib/api/typesWithBodyKey";
 import { logger } from "@/common/logger";
 import { CollectionsEntity } from "@/models/collections/collections-entity";
 import { ActivityDocument } from "@/elasticsearch/indexes/activities/base";
+import { getNetworkName } from "@/config/network";
+import { config } from "@/config/index";
 
-const INDEX_NAME = "activities";
+const INDEX_NAME = `${getNetworkName()}.activities`;
 
 const MAPPINGS: MappingTypeMapping = {
   dynamic: "false",
@@ -223,6 +225,13 @@ export const createIndex = async (): Promise<void> => {
         },
         index: `${INDEX_NAME}-${Date.now()}`,
         mappings: MAPPINGS,
+        settings: {
+          number_of_shards: config.chainId === 5 ? 4 : 40,
+          sort: {
+            field: ["timestamp", "createdAt"],
+            order: ["desc", "desc"],
+          },
+        },
       });
 
       if (elasticsearchCloud) {
--- a/packages/indexer/src/jobs/elasticsearch/backfill-ask-activities-elasticsearch.ts
+++ b/packages/indexer/src/jobs/elasticsearch/backfill-ask-activities-elasticsearch.ts
@@ -31,6 +31,9 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
     async (job: Job) => {
       const cursor = job.data.cursor as CursorInfo;
 
+      const fromTimestamp = job.data.fromTimestamp || 0;
+      const toTimestamp = job.data.toTimestamp || 9999999999;
+
       const limit = Number((await redis.get(`${QUEUE_NAME}-limit`)) || 1);
 
       try {
@@ -43,6 +46,7 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
         const query = `
             ${AskCreatedEventHandler.buildBaseQuery()}
             WHERE side = 'sell'
+            AND (updated_at >= to_timestamp($/fromTimestamp/) AND updated_at < to_timestamp($/toTimestamp/)) 
             ${continuationFilter}
             ORDER BY updated_at, id
             LIMIT $/limit/;
@@ -51,6 +55,8 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
         const results = await ridb.manyOrNone(query, {
           id: cursor?.id,
           updatedAt: cursor?.updatedAt,
+          fromTimestamp,
+          toTimestamp,
           limit,
         });
 
@@ -70,14 +76,21 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
 
           await ActivitiesIndex.save(activities);
 
-          logger.info(QUEUE_NAME, `Processed ${results.length} activities.`);
-
           const lastResult = results[results.length - 1];
 
-          await addToQueue({
-            updatedAt: lastResult.updated_ts,
-            id: lastResult.order_id,
-          });
+          logger.info(
+            QUEUE_NAME,
+            `Processed ${results.length} activities. fromTimestamp=${fromTimestamp}, toTimestamp=${toTimestamp}, lastTimestamp=${lastResult.event_timestamp}`
+          );
+
+          await addToQueue(
+            {
+              updatedAt: lastResult.updated_ts,
+              id: lastResult.order_id,
+            },
+            fromTimestamp,
+            toTimestamp
+          );
         }
       } catch (error) {
         logger.error(
@@ -96,17 +109,24 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
   });
 
   redlock
-    .acquire([`${QUEUE_NAME}-lock-v9`], 60 * 60 * 24 * 30 * 1000)
+    .acquire([`${QUEUE_NAME}-lock-v13`], 60 * 60 * 24 * 30 * 1000)
     .then(async () => {
-      await addToQueue();
+      await addToQueue(undefined, undefined, 1609459199);
+      await addToQueue(undefined, 1609459200, 1640995199);
+      await addToQueue(undefined, 1640995200, 1672531199);
+      await addToQueue(undefined, 1672531200);
     })
     .catch(() => {
       // Skip on any errors
     });
 }
 
-export const addToQueue = async (cursor?: CursorInfo) => {
-  await queue.add(randomUUID(), { cursor });
+export const addToQueue = async (
+  cursor?: CursorInfo,
+  fromTimestamp?: number,
+  toTimestamp?: number
+) => {
+  await queue.add(randomUUID(), { cursor, fromTimestamp, toTimestamp });
 };
 
 export interface CursorInfo {
--- a/packages/indexer/src/jobs/elasticsearch/backfill-bid-activities-elasticsearch.ts
+++ b/packages/indexer/src/jobs/elasticsearch/backfill-bid-activities-elasticsearch.ts
@@ -32,6 +32,9 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
     async (job: Job) => {
       const cursor = job.data.cursor as CursorInfo;
 
+      const fromTimestamp = job.data.fromTimestamp || 0;
+      const toTimestamp = job.data.toTimestamp || 9999999999;
+
       const limit = Number((await redis.get(`${QUEUE_NAME}-limit`)) || 1000);
 
       try {
@@ -45,6 +48,7 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
             ${BidCreatedEventHandler.buildBaseQuery()}
             WHERE side = 'buy'
             AND fillability_status = 'fillable' AND approval_status = 'approved'
+            AND (updated_at >= to_timestamp($/fromTimestamp/) AND updated_at < to_timestamp($/toTimestamp/)) 
             ${continuationFilter}
             ORDER BY updated_at, id
             LIMIT $/limit/;
@@ -53,6 +57,8 @@ if (config.doBackgroundWork && config.doElasticsearchWork) {
         const results = await ridb.manyOrNone(query, {
           id: cursor?.id,
           updatedAt: cursor?.updatedAt,
+          fromTimestamp,
+          toTimestamp,
           limit,
         });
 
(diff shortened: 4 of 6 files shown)


lljxx1:
looks good to me

packages/contracts/contracts/router/modules/exchanges/CollectionXyzModule.sol
```diff
@@ -0,0 +1,220 @@
...
+    external
+    payable
+    nonReentrant
+    refundETHLeftover(params.refundTo)
+    chargeETHFees(fees, params.amount)
+  {
+    uint256 poolsLength = pools.length;
+    for (uint256 i; i < poolsLength; ) {
+      // Build router data
+      ICollectionRouter.PoolSwapSpecific[]
```
georgeroman, line 55: Would it be more gas-efficient to declare this outside of the `for` loop? Same applies below to `buyWithERC20`.

packages/contracts/contracts/router/modules/exchanges/CollectionXyzModule.sol
```diff
@@ -0,0 +1,220 @@
...
+        pool: pools[i],
+        nftIds: new uint256[](1),
+        proofs: orderParams[i].proofs,
+        proofFlags: orderParams[i].proofFlags,
+        externalFilterContext: orderParams[i].externalFilterContext
+      });
+      swapList[0].nftIds[0] = orderParams[i].nftId;
+
+      // Fetch the current price quote
+      (, , uint256 price, ) = pools[i].getBuyNFTQuote(1);
```
georgeroman, line 67: Could this fail if there isn't enough liquidity in the pool? Might be worth putting this in the `try / catch` below.

packages/indexer/src/migrations/1682410320474_collection-pools.sql
```diff
@@ -0,0 +1,20 @@
+-- Up Migration
+
+CREATE TABLE "collection_pools" (
```
georgeroman, line 3: Let's rename this to `collectionxyz_pools`.

packages/indexer/src/models/sources/sources.json
```diff
@@ -63,6 +63,17 @@
       "tokenUrlGoerli": "https://knny.vercel.app/${contract}/${tokenId}"
     }
   },
+  "7": {
```
georgeroman, line 66: This is not needed anymore, your source will automatically get inserted once a new order comes through.

packages/sdk/src/collectionxyz/addresses.ts
```diff
@@ -0,0 +1,51 @@
+import { ChainIdToAddress, Network } from "../utils";
+
+export const CollectionPoolEnumerableETH: ChainIdToAddress = {
```
georgeroman, line 3: Addresses should all be lowercase.

packages/sdk/src/router/v5/router.ts
```diff
@@ -393,6 +393,15 @@ export class Router {
       };
     }
 
+    // The V5 router does not support filling CollectionXyz bids, so we fill directly
```
georgeroman, line 396: The v5 router is deprecated so this is not needed.

packages/sdk/src/router/v6/router.ts
```diff
@@ -357,6 +363,12 @@ export class Router {
       }
     }
 
+    if (details.some(({ kind }) => kind === "blur")) {
```
georgeroman, line 366: This was added by mistake?

packages/sdk/src/router/v6/types.ts
```diff
@@ -95,6 +95,10 @@ export type GenericOrder =
       kind: "sudoswap";
       order: Sdk.Sudoswap.Order;
     }
+  | {
+      kind: "collection";
```
georgeroman, line 99: Let's rename this to "collectionxyz"

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -0,0 +1,938 @@
...
+    price: currencyPrice,
+    value: currencyValue,
+    normalizedValue: currencyNormalizedValue,
+  };
+};
+
+/**
+ * Get the address which the pool would pay royalties to for `tokenId`
+ */
+const getRoyaltyRecipient = async (
```
georgeroman, line 165: We already have this information, no need to make an on-chain call for every order:
https://github.com/reservoirprotocol/indexer/blob/bfc979efa63f3e5b5847de61d936158006b8858a/packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts#L190-L198

georgeroman (changes requested):

packages/contracts/contracts/router/modules/exchanges/CollectionXyzModule.sol
potato-cat69: Nope, quote functions dont revert if the pool cant take it (this is intentional so users can know exactly how much they might need to top up to be able to buy more nfts with the pool)

packages/indexer/docker-compose.yaml
```diff
@@ -3,6 +3,7 @@ version: "3"
 services:
   postgres:
     image: postgres:13.2
+    command: postgres -N 2000
```
georgeroman, line 6: Just curious, had any issues running without `-N 2000`?

packages/sdk/src/router/v6/router.ts
```diff
@@ -3360,6 +3451,42 @@ export class Router {
           break;
         }
 
+        case "collectionxyz": {
+          const order = detail.order as Sdk.CollectionXyz.Order;
+          const module = this.contracts.collectionXyzModule;
+          const acceptedSet = detail.order.params.acceptedSet.map(BigInt);
```
georgeroman, line 3457: As noted above, should pass the token ids via an additional field rather than assuming they are stored in the raw order data itself.

packages/indexer/src/sync/events/handlers/collectionxyz.ts
```diff
@@ -0,0 +1,454 @@
...
+              isModifierEvent: true,
+            },
+            metadata: {},
+          },
+        });
+
+        const parsedLog = eventData.abi.parseLog(log);
+        const pool = await getPoolDetails(baseEventParams.address);
+
+        const txTrace = await utils.fetchTransactionTrace(txHash);
```
georgeroman, line 69: For both `swap-nft-out-pool` and `swap-nft-in-pool` I think we can use the event data rather than fetching the internal call trace. The reason we use tracing for sudoswap is because their events are badly designed and don't include any information:
https://github.com/reservoirprotocol/indexer/blob/92a7cca1be58b0f24578cb06a337aa9a4de0f7f1/packages/indexer/src/sync/events/data/sudoswap.ts#L7-L21

But in your case the events seem to include all data that we need. Tracing is also a bit more expensive, slower and more prone to weird issues, so would rather avoid it unless strictly necessary.

georgeroman (changes requested):

packages/indexer/docker-compose.yaml
potato-cat69: We would frequently get "Sorry too many connections"

packages/sdk/src/collectionxyz/types.ts
```diff
@@ -0,0 +1,14 @@
+export type OrderParams = {
+  pool: string;
+  // Only relevant for listings
+  tokenId?: string;
+  externalFilter: string;
+  // Set of tokenIds accepted by the merkle root of this pool. [] for buy orders
+  // (buy from pool) as proofs aren't needed when buying, or sell orders for
+  // unfiltered pools
+  acceptedSet: string[];
```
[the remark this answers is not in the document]
potato-cat69: ok, in that case ill also update the indexer to pass over bid orders for sets >100k. btw is it strictly or weakly greater than 100k? We saw some issues saving otherdeed orders on testnet which has 100k token ids exactly

packages/indexer/src/sync/events/handlers/collectionxyz.ts
```diff
@@ -0,0 +1,454 @@
...
+        trades.sell.set(`${txHash}-${address}`, tradeRank + 1);
+
+        break;
+      }
+
+      case "collectionxyz-new-pool": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const pool = parsedLog.args["poolAddress"].toLowerCase();
+
+        // New pools will need encoded tokenIds from the AcceptsTokenIDs event
```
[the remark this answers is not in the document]
potato-cat69: > // New pools will need encoded tokenIds from the AcceptsTokenIDs event

This step? unfortunately no, we dont store the token ids on chain, would be too expensive. its only emitted once via the creation/update events

packages/indexer/src/sync/events/data/collectionxyz.ts
```diff
@@ -0,0 +1,267 @@
...
+import { EventData } from "@/events-sync/data";
+
+export const newPool: EventData = {
+  kind: "collectionxyz",
+  subKind: "collectionxyz-new-pool",
+  topic: "0x77948cb83ef3caff9ac13dfab1ea1f8a6875c98370287ce587f5dbc74cc5b6b0",
+  numTopics: 3,
+  abi: new Interface([
+    `event NewPool(
+      address indexed collectionxyz,
```
georgeroman, line 12: I assume the find and replace also updated `collection` to `collectionxyz` here, probably makes sense to keep as `collection` here and below.

packages/indexer/src/sync/events/handlers/collectionxyz.ts
```diff
@@ -0,0 +1,454 @@
...
+              return acceptsTokenIds.abi.parseLog(log);
+            } catch (err) {
+              return undefined;
+            }
+          })
+          .filter((log) => log !== undefined)[0];
+
+        // encodedTokenIds is [] to represent unfiltered pool. undefined value
+        // is reserved for events which don't modify encodedTokenIds
+        const encodedTokenIds = hexToBytes(acceptsTokenIdsLog?.args["_data"] ?? "0x");
```
georgeroman, line 377: As mentioned above, let's pass this as hex and only convert in the `save` method.

packages/indexer/src/sync/events/handlers/collectionxyz.ts
```diff
@@ -0,0 +1,454 @@
...
+            },
+            metadata: {},
+          },
+        });
+
+        break;
+      }
+      case "collectionxyz-accepts-token-ids": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const encodedTokenIds = hexToBytes(parsedLog.args["_data"] ?? "0x");
```
georgeroman, line 399: Same here

georgeroman (changes requested):

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -0,0 +1,961 @@
...
+} from "@/models/collection-pools";
+import { BigNumber, BigNumberish, ethers } from "ethers";
+import { TokenIDs } from "fummpel";
+import { getUSDAndNativePrices } from "@/utils/prices";
+import MerkleTree from "merkletreejs";
+import { keccak256 as keccakWithoutTypes } from "@ethersproject/keccak256";
+
+const hashFn = (tokenId: BigNumberish) => keccak256(["uint256"], [tokenId]);
+
+const generateMerkleTree = (tokenIds: BigNumberish[]) => {
```
georgeroman, line 31: Any reason not to reuse the code we already have for the merkle roots: https://github.com/reservoirprotocol/indexer/blob/main/packages/sdk/src/common/helpers/merkle.ts?

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -0,0 +1,961 @@
...
+    recipient: string;
+    bps: number;
+  }[];
+  totalFeeBps: number;
+  royaltyBps: number;
+  protocolBps: number;
+  tradeBps: number;
+  carryBps: number;
+}> => {
+  const [tradeBps, protocolBps, royaltyBps, carryBps] = (await poolContract.feeMultipliers()).map(
```
georgeroman, line 87: It's a bit redundant to make an on-chain call for every order. What about storing all the different fees in the `collectionxyx_pools` table in a `metadata JSONB` column. This way we would save these fees only the first time we encounter the pool (also when the fees get updated we'll have to refresh them). Is that viable?

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -0,0 +1,961 @@
...
+      "onchain"
+    );
+    // Assume ERC2981 if and only if 1 recipient
+    if (onChainRoyalties.length !== 1) throw new Error();
+    erc2981Recipient = onChainRoyalties[0].recipient;
+  } catch {
+    // Leave as address(0)
+  }
+
+  return poolContract.getRoyaltyRecipient(erc2981Recipient);
```
georgeroman, line 184: What's this additional call needed for? Maybe we can cache it somehow.

georgeroman (changes requested):

packages/indexer/src/sync/events/data/collectionxyz.ts
potato-cat69: Yes, thanks for the catch

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
potato-cat69: importing that function causes a rather mysterious bug as shown in the ss below. Avoiding the import was as specific a fix as we could find

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
potato-cat69: Sure, I'll add a flag for fee updates to the orderinfo based on the event trigger

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
georgeroman: Hmm, was it a memory leak issue when starting the indexer? Make sure to import from `@reservoir0x/sdk` rather than `@reservoir/sdk/src`. Anyway, not a big deal.

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -0,0 +1,1089 @@
...
+          `function getRoyaltyRecipient(address payable erc2981Recipient) view returns (address payable)`,
+          `function getAllHeldIds() view returns (uint256[])`,
+          `function tokenIDFilterRoot() view returns (bytes32)`,
+          `function externalFilter() public view returns (address)`,
+        ]),
+        baseProvider
+      );
+
+      const isERC20 = pool.token !== Sdk.Common.Addresses.Eth[config.chainId];
+      const externalFilterAddress = await poolContract.externalFilter();
```
georgeroman, line 384: Can we get this from the `collectionxyz_pools` table instead of making an on-chain call?

georgeroman (changes requested):

packages/indexer/src/orderbook/orders/collectionxyz/index.ts
potato-cat69: > Can we get this from the collectionxyz_pools table instead of making an on-chain call?

I don't think that would be the best approach because that table currently holds immutable params only; if we added externalFilter (a mutable param) to that table, we'd need to add separate logic for external filter updates. i'll put it in the raw_data column similar to the previous fixes

potato-cat69: >Hmm, was it a memory leak issue when starting the indexer? Make sure to import from @reservoir0x/sdk rather than @reservoir/sdk/src. Anyway, not a big deal.

Not too sure, it oom's without logs so it's hard for us to trace with minimal architectural understanding. If you run the code locally with that fn imported it might be easier for you to see what's going on under the hood

PR #4431, opened 2023-05-05
fix: jamie/ops-356-explain-the-kinds-of-bids-in-ordersbids

jamiechicago312:
* Does not show up correctly Changing to ª to show correctly in docs


jamiechicago312:
Changing symbols on /orders/asks and /orders/bids

jamiechicago312:
Will expand bids description to explain types of bids returned

nofir (approved):

PR #4440, opened 2023-05-06
Update network.ts

westonnelson:
Add $ARB and $NFTE on Arbitrum as currencies to be enabled for use on the marketplace.

### DIFF
--- a/packages/indexer/src/config/network.ts
+++ b/packages/indexer/src/config/network.ts
@@ -490,9 +490,9 @@ export const getNetworkSettings = (): NetworkSettings => {
         subDomain: "api-arbitrum",
         whitelistedCurrencies new Map ([
           [
-            "0x912CE59144191C1204E64559FE8253a0e49E6548", // MUST BE LOWERCASE
+            "0x912ce59144191c1204e64559fe8253a0e49e6548", // MUST BE LOWERCASE
             {
-              contract: "0x912CE59144191C1204E64559FE8253a0e49E6548",
+              contract: "0x912ce59144191c1204e64559fe8253a0e49e6548",
               name: "Arbitrum",
               symbol: “ARB”,
               decimals: 18,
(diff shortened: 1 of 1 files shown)


packages/indexer/src/config/network.ts
```diff
@@ -488,6 +488,34 @@ export const getNetworkSettings = (): NetworkSettings => {
         lastBlockLatency: 5,
         headBlockDelay: 10,
         subDomain: "api-arbitrum",
+        whitelistedCurrencies new Map ([
+          [
+            "0x912CE59144191C1204E64559FE8253a0e49E6548", // MUST BE LOWERCASE
```
nofir, line 493: can you make this lowercase?

nofir (changes requested):

packages/indexer/src/config/network.ts
westonnelson: > can you make this lowercase?

Yes, done!

nofir (approved):

nofir (dismissed):

packages/indexer/src/config/network.ts
nofir: @westonnelson the whitelisted currencies are only for currencies that are not supported by coingecko (which we use to keep currencies conversions up to date). I see both currencies are available on coingecko and already supported on our Arbitrum indexer.

nofir: @westonnelson just to confirm, we can close this PR? 

You can see the currencies in coingecko here:
- https://www.coingecko.com/en/coins/arbitrum
- https://www.coingecko.com/en/coins/nftearth

westonnelson: Hi there! - the challenge is that I see the currencies when we're using the Reservoir Indexer; but when we were using our own implementation of the Indexer and using Alchemy/GCP we could actually list the items in $ARB and $NFTE and now it is not functional.

There appears to be some sort of disconnect here; I do see seem as you mention but can't actually list when using Reservoir Indexer so thought this would be the solve. 

Wondering now what the root issue may be.

Screencaps attached for your review.

pedromcunha: Hi, this is likely an issue for ReservoirKit. Can you open up an issue in this repo so we can troubleshoot: https://github.com/reservoirprotocol/reservoir-kit

nofir:
Closing this for now

PR #4477, opened 2023-05-10
Support mantle testnet

xcaptain:
The official website: https://www.mantle.xyz/


packages/indexer/src/config/network.ts
```diff
@@ -553,6 +556,39 @@ export const getNetworkSettings = (): NetworkSettings => {
...
     }
+    case 5001: {
+      return {
+        ...defaultNetworkSettings,
+        enableWebSocket: false,
+        realtimeSyncMaxBlockLag: 32,
+        realtimeSyncFrequencySeconds: 5,
+        lastBlockLatency: 5,
+        headBlockDelay: 10,
+        subDomain: "api-scroll-alpha",
```
georgeroman, line 567: Shouldn't this be a different subdomain?

xcaptain: Thanks, fixed

PR #4513, opened 2023-05-12
Feat/activities batches

nofir:


packages/indexer/src/jobs/activities/utils.ts
```diff
@@ -59,6 +59,60 @@ export async function getBidInfoByOrderId(orderId: string) {
...
+                SELECT
+                  orders.id AS "order_id",
+                  ts.id,
+                  ts.attribute_id,
+                  ts.collection_id
+                FROM orders
+                JOIN token_sets ts
+                  ON orders.token_set_id = ts.id
+                WHERE orders.id IN ($/orderIds:list/)
+                LIMIT 1
```
nofir, line 78: isnt the LIMIT 1 an issue?

ipeleg: Right, too much copy paste

PR #4570, opened 2023-05-16
feat: response api descriptions for collection endpoints

jamiechicago312:
Created a variety of collection endpoint response descriptions.

### DIFF
--- a/packages/indexer/src/api/endpoints/collections/get-user-collections/v3.ts
+++ b/packages/indexer/src/api/endpoints/collections/get-user-collections/v3.ts
@@ -125,8 +125,8 @@ export const getUserCollectionsV3Options: RouteOptions = {
             }.description("The floor sale from X-days ago."),
           }),
           ownership: Joi.object({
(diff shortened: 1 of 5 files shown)


packages/indexer/src/api/endpoints/collections/get-user-collections/v3.ts
```diff
@@ -106,27 +106,27 @@ export const getUserCollectionsV3Options: RouteOptions = {
...
               "1day": Joi.number().unsafe().allow(null),
               "7day": Joi.number().unsafe().allow(null),
               "30day": Joi.number().unsafe().allow(null),
-            },
+            }.description("The floor sale from X-days ago."),
           }),
           ownership: Joi.object({
-            tokenCount: Joi.string(),
-            onSaleCount: Joi.string(),
+            tokenCount: Joi.string().description("Owned tokens from this collection"),
```
nofir, line 128: "Owned"?

isn't a token always owned?

PR #4578, opened 2023-05-17
feat: process bids via queue

ipeleg:


packages/indexer/src/websockets/opensea/index.ts
```diff
@@ -97,7 +100,14 @@ if (config.doWebsocketWork && config.openSeaApiKey) {
               const startTime = now();
               if (bidsEvents.length >= maxEventsSize) {
                 const orderInfoBatch = bidsEvents.splice(0, bidsEvents.length);
-                await orderbookOrders.addToQueue(orderInfoBatch);
```
nofir, line 100: it should be only for single token bids. and maybe we can add the trait offers too.

nofir (changes requested):

nofir (approved):

PR #4582, opened 2023-05-17
Tv/platf 1211 attribute issues with large collection

tv3636:


packages/indexer/src/api/endpoints/attributes/get-attributes-explore/v4.ts
```diff
@@ -82,7 +82,7 @@ export const getAttributesExploreV4Options: RouteOptions = {
           tokenCount: Joi.number().required(),
           onSaleCount: Joi.number().required(),
           sampleImages: Joi.array().items(Joi.string().allow("", null)),
-          floorAskPrices: Joi.array().items(Joi.number().unsafe()),
+          floorAskPrices: Joi.array().items(Joi.number().unsafe().allow(null)),
```
nofir, line 85: I think it might need to be a new version given that now we return null so we're basically changing the response or we can either return 0 which is like what the behavior now when there is no floor ask (might be a bit confusing).

wdyt?

tv3636: I think it's best to return null rather than 0 to avoid confusion. I'll move it to a new version to prevent breaking anyone who might be relying on numeric responses

packages/indexer/src/api/endpoints/attributes/get-attributes-explore/v5.ts
```diff
@@ -0,0 +1,299 @@
...
+  response: {
+    schema: Joi.object({
+      attributes: Joi.array().items(
+        Joi.object({
+          key: Joi.string().required(),
+          value: JoiAttributeValue,
+          tokenCount: Joi.number().required(),
+          onSaleCount: Joi.number().required(),
+          sampleImages: Joi.array().items(Joi.string().allow("", null)),
+          floorAskPrices: Joi.array().items(Joi.number().unsafe().allow(null)),
```
nofir, line 85: aren't you missing the description?

nofir (changes requested):

packages/indexer/src/api/endpoints/attributes/get-attributes-explore/v5.ts
tv3636: the description is updated on [line 22](https://github.com/reservoirprotocol/indexer/pull/4582/files#diff-e5a252d272c243b8fc4d21bbfb5221b63180999171e7c052e20a5fcce9f14c87R22) - not sure what you mean, can a description be added here?

packages/indexer/src/api/endpoints/attributes/get-attributes-explore/v5.ts
```diff
@@ -0,0 +1,299 @@
...
+        value: r.value,
+        tokenCount: Number(r.token_count),
+        onSaleCount: Number(r.on_sale_count),
+        sampleImages: Assets.getLocalAssetsLink(r.sample_images) || [],
+        floorAskPrices:
+          Number(r.token_count) <= 10000 // We only calculate attribute floor up to 10k tokens
+            ? query.maxFloorAskPrices > 1
+              ? (r.floor_sell_values || []).map(formatEth)
+              : [formatEth(r.floor_sell_value || 0)]
+            : [null],
```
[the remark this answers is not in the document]
tv3636: updated to return an empty array

packages/indexer/src/api/endpoints/attributes/get-attributes-explore/v5.ts
nofir: I see. i thought maybe before it was on the property. any reason not to put it there?

tv3636: [On attributes/all it's in the main description](https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/api/endpoints/attributes/get-attributes-all/v4.ts#L19) like I have here. Do you mean putting it on `maxFloorAskPrices`? It seems confusing to put there since that's for setting a separate limit

nofir (approved):

PR #4591, opened 2023-05-18
Request support for $NELK token

dturska12:
reservoirprotocol/backend team


dturska12:
Request support for $NELK token

rorygeoghegan:
@dturska12 this PR is for Pepe and not $NELK token. Pepe which is already supported.

PR #4599, opened 2023-05-18
feat: add extra image metadata to tokens table and tokens endpoints

ted-palmer:
This pull request is dependent on [this PR](https://github.com/reservoirprotocol/metadata/pull/136) from the metadata api


packages/indexer/src/utils/assets.ts
```diff
@@ -42,4 +42,25 @@ export class Assets {
...
+    if (imageUrl?.includes("i.seadn.io")) {
+      if (imageUrl.match(/w=\d+/)) {
+        return imageUrl.replace(/w=\d+/, `w=${size}`);
+      } else {
+        return `${imageUrl}?w=${size}`;
+      }
+    }
+
+    // If the image provider is not recognized, return undefined
+    return undefined;
```
nofir, line 64: I think undefined is returned by default.

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -204,6 +204,9 @@ export const getTokensV6Options: RouteOptions = {
             name: Joi.string().allow("", null),
             description: Joi.string().allow("", null),
             image: Joi.string().allow("", null),
+            imageSmallUrl: Joi.string().allow("", null),
```
nofir, line 207: very small comment.. maybe just to keep consistent: imageSmall /imageLarge (unless peter asked specifically)

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1055,6 +1059,11 @@ export const getTokensV6Options: RouteOptions = {
             name: r.name,
             description: r.description,
             image: Assets.getLocalAssetsLink(r.image),
+            imageSmallUrl: Assets.getResizedImageUrl(r.image, 250),
+            imageLargeUrl: Assets.getResizedImageUrl(r.image, 1000),
+            metadata: {
```
nofir, line 1064: was the requirement to return the original in metadata?

maybe check if image_original_url exists and then decide if to return the metdata:

metadata: r.metadata.image_original_url ? { ... } : undefined

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1055,6 +1059,11 @@ export const getTokensV6Options: RouteOptions = {
             name: r.name,
             description: r.description,
             image: Assets.getLocalAssetsLink(r.image),
+            imageSmallUrl: Assets.getResizedImageUrl(r.image, 250),
```
nofir, line 1062: maybe use ImageSize.small / ImageSize.large?

packages/indexer/src/migrations/1643714286947_tokens.sql
```diff
@@ -6,6 +6,7 @@ CREATE TABLE "tokens" (
   "name" TEXT,
   "description" TEXT,
   "image" TEXT,
+  "metadata" JSONB,
```
nofir, line 9: this is wrong, you should create a new migration with metdata, otherwise this migration will never run

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
ted-palmer: Will do!

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
ted-palmer: I think Peter wanted these as top level items because we intend to always return them. Is your suggestion pretty much to have an ImageSize object like so?:

```
ImageSize: {
        small: Assets.getResizedImageUrl(r.image, 250),
        large: Assets.getResizedImageUrl(r.image,1000),
}
```

packages/indexer/src/migrations/1643714286947_tokens.sql
ted-palmer: Ah ok, didn't realize that's how it worked. How should I go about creating a new migration? I see the filenames have a some sort of unique number at the beginning

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
nofir: No, its to use the enum you created so if we change the size, you dont need to change everywhere. (you might want to move it to the assets class
```
enum ImageSize {
  small = 250,
  medium = 512,
  large = 1000,
}
```

packages/indexer/src/migrations/1643714286947_tokens.sql
nofir: create a new file with the current timestamp in milliseconds and then add the field.

see here: https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/migrations/1684443851969_collections-owners-count.sql

nofir (changes requested):
looks good, left one comment

packages/indexer/src/migrations/1685050927714_tokens-metadata.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+
+ALTER TABLE "tokens" ADD COLUMN "metadata" JSONB;
+
+-- Down Migration
```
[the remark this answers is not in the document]
ted-palmer: Yep! Done ✅

nofir (approved):

PR #4613, opened 2023-05-19
feat: support blend

lljxx1:


packages/indexer/calls.json
```diff
@@ -0,0 +1,102 @@
+{
```
georgeroman, line 1: Let's remove this file.

packages/indexer/src/sync/events/handlers/blend.ts
```diff
@@ -0,0 +1,144 @@
...
+          { to: exchangeAddress, type: "CALL", sigHashes: ["0xe7efc178"] },
+          tradeRank
+        );
+
+        if (!executeCallTrace) {
+          break;
+        }
+
+        const inputData = exchange.contract.interface.decodeFunctionData(
+          "buyLocked",
```
georgeroman, line 55: There's multiple method call that could result in the `BuyLocked` event being emitted:
- `buyLocked`
- `buyLockedETH`
- `buyToBorrowLocked`
- `buyToBorrowLockedETH`
We should cover all these cases.

packages/indexer/src/sync/events/handlers/blend.ts
```diff
@@ -0,0 +1,144 @@
...
+          nonce: minNonce.toString(),
+          signature: inputData.signature,
+        });
+
+        let isValidated = true;
+        const orderId = order.hash();
+        try {
+          order.checkSignature();
+        } catch {
+          isValidated = false;
```
georgeroman, line 96: Instead of failing directly, we should probably try every nonce:
```
for (let nonce = minNonce; nonce >= 0; nonce--) {
  order.nonce = nonce;
  ...
}
```

packages/indexer/src/sync/events/handlers/blend.ts
```diff
@@ -0,0 +1,178 @@
...
+          }
+        }
+
+        if (!isValidated) {
+          // not validated
+          return;
+        }
+
+        // Handle: attribution
+        const orderKind = "blur";
```
georgeroman, line 139: This is a bit misleading, let's set the order kind to 'blend' and make sure to map it to the `blur.io` source:https://github.com/reservoirprotocol/indexer/blob/96e1d3062ca46b575b525c27ffb872a1777a2071/packages/indexer/src/orderbook/orders/index.ts#L157-L158

PR #4643, opened 2023-05-22
fix: nft length for price calc when sell to pool

potato-cat69:


packages/indexer/src/sync/events/handlers/collectionxyz.ts
```diff
@@ -262,7 +262,7 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
 
             let taker = decodedInput.tokenRecipient.toLowerCase();
             const price = bn(parsedLog.args["outputAmount"])
-              .div(decodedInput.nfts.length)
+              .div(decodedInput.nfts.ids.length)
```
georgeroman, line 265: Can we do the same for `swap-nfts-out-pool`?

PR #4668, opened 2023-05-23
Development

potato-cat69:


packages/indexer/src/orderbook/orders/collectionxyz/index.ts
```diff
@@ -430,18 +430,23 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
             // Determine how many NFTs can be bought (though the price will
             // increase with each unit)
             let numBuyableNFTs = 1;
+            const pricesAsBn: BigNumber[] = [];
             // Hardcoded limit in case there's way too much liquidity
             while (numBuyableNFTs < 10) {
               const { totalAmount }: { totalAmount: BigNumber } =
                 await poolContract.getSellNFTQuote(numBuyableNFTs);
+              const unitPrice = totalAmount.sub(pricesAsBn[0] ?? bn(0));
```
georgeroman, line 438: Should we subtract `pricesAsBn[pricesAsBn.length - 1]` instead of `pricesAsBn[0]`? Same comment for below changes.

potato-cat69: Right, thanks for the catch

PR #4726, opened 2023-05-25
Added WEN Token to Whitelist

zorpheux:


nofir:
@zorpheux We retrieve currencies from coingecko and it seems its already supported: https://www.coingecko.com/en/coins/wen-token

so no need to add custom handling.

PR #4758, opened 2023-05-26
feat: adding sync node logic to bids api endpoint

JadenDurnford:


packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
```diff
@@ -309,29 +328,94 @@ export const getOrdersBidsV5Options: RouteOptions = {
...
+      ) {
+        throw Boom.badRequest(
+          `You must provide one of the following: [ids, maker, contracts] in order to filter querys with sortBy = updatedAt and status != 'active.`
+        );
+      }
+
+      // TODO Remove this restriction once an index is created for updatedAt and contracts
+      if (query.sortBy === "updatedAt" && query.contracts && query.status === "any") {
+        throw Boom.badRequest(
+          `Cannot filter by contracts while sortBy = "updatedAt" and status = "any"`
```
nofir, line 386: isnt the whole point is to be able to get all bids if sorting by update at?

JadenDurnford: Yeah, I was running into some issues with the query timing out when adding the sorting by contracts as well, so we might need to add an index to enable it.

packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
```diff
@@ -309,29 +328,94 @@ export const getOrdersBidsV5Options: RouteOptions = {
...
+        query.status !== "active" &&
+        (query.token ||
+          query.tokenSetId ||
+          query.community ||
+          query.collectionsSetId ||
+          query.native ||
+          query.source)
+      ) {
+        throw Boom.badRequest(
+          `You must provide one of the following: [ids, maker, contracts] in order to filter querys with sortBy = updatedAt and status != 'active.`
```
nofir, line 379: I actually meant to put the comment below here: "isnt the whole point is to be able to get all bids if sorting by update at?"?

nofir (dismissed):

PR #4771, opened 2023-05-29
feat: sudoswap v2

lljxx1:


packages/contracts/contracts/router/modules/exchanges/SudoswapV2Module.sol
```diff
@@ -0,0 +1,188 @@
...
+
+      unchecked {
+        ++i;
+      }
+    }
+  }
+
+  // --- Single ERC721 offer ---
+
+  function sell(
```
georgeroman, line 117: Sudoswap V2 also supports ERC1155, which we should cover.

packages/contracts/test/router/helpers/sudoswap-v2.ts
```diff
@@ -0,0 +1,137 @@
...
+import { AddressZero } from "@ethersproject/constants";
+import { bn, getChainId } from "../../utils";
+
+import FactoryAbi from "@reservoir0x/sdk/src/sudoswap-v2/abis/Factory.json";
+
+// --- Listings ---
+
+export type SudoswapListing = {
+  seller: SignerWithAddress;
+  nft: {
```
georgeroman, line 15: We should test ERC1155s as well, see below for an example: https://github.com/reservoirprotocol/indexer/blob/2cc158452b81979ea4f38ea8daf977fc534cc6ac/packages/contracts/test/router/helpers/zeroex-v4.ts#L14

packages/indexer/src/migrations/1685063598297_sudoswap-v2-pools.sql
```diff
@@ -0,0 +1,18 @@
+-- Up Migration
+
+CREATE TABLE "sudoswap_v2_pools" (
```
georgeroman, line 3: We should also keep track of the `nft_kind` here (ERC721 vs ERC1155)

georgeroman (changes requested):

PR #4824, opened 2023-05-31
feat: add enableRealtimeV2BlockQueue & realtimeV2ChainIds

d3or:


packages/indexer/src/config/network.ts
```diff
@@ -130,6 +130,8 @@ type NetworkSettings = {
   };
   onStartup?: () => Promise<void>;
   subDomain: string;
+  enableRealtimeV2BlockQueue: boolean;
```
nofir, line 133: Aren't the changes here redundant if you are actually using the env var from the config?

nofir (approved):

PR #4867, opened 2023-06-02
deor/platf 1260 refactor back fill

d3or:
- feat: slight improvements to backfill
- feat: subset log processing instead of bulk


ipeleg (approved):
Looks good, let's deploy on Monday

PR #4902, opened 2023-06-06
feat: temp disable looksrare-v2 fee

lljxx1:


packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -176,13 +176,11 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
...
 
       // Handle: fees
-      let feeBreakdown = [
-        {
-          kind: "marketplace",
-          recipient: "0x5924a28caaf1cc016617874a2f0c3710d881f3c1",
-          bps: 50,
-        },
-      ];
+      let feeBreakdown: {
```
georgeroman, line 179: We should:
- change marketplace fee recipient
- remove royalty fees

image

georgeroman (changes requested):

PR #5044, opened 2023-06-09
Tv/platf 1398 store opensea collection payment tokens

tv3636:


packages/indexer/src/api/endpoints/collections/get-collection-supported-marketplaces/v2.ts
```diff
@@ -0,0 +1,294 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
nofir, line 1: Why is there a need for a v2 if we're only adding paymentTokens?

packages/indexer/src/api/endpoints/collections/get-collection-supported-marketplaces/v2.ts
```diff
@@ -0,0 +1,294 @@
...
+                maxBps: maxOpenseaRoyaltiesBps,
+              }
+            : undefined,
+          orderbook: "opensea",
+          orderKind: "seaport-v1.4",
+          listingEnabled: false,
+          customFeesSupported: false,
+          minimumBidExpiry: 15 * 60,
+          supportedBidCurrencies: Object.keys(ns.supportedBidCurrencies),
+          paymentTokens: collectionResult.payment_tokens?.opensea,
```
nofir, line 220: Are you planning to do a backfill? alternatively you can do something similar to line 186 and try to get on the fly

packages/indexer/src/jobs/token-updates/fetch-collection-metadata-job.ts
```diff
@@ -88,6 +90,7 @@ export class FetchCollectionMetadataJob extends AbstractRabbitMqJobHandler {
           tokenIdRange,
           tokenSetId: collection.tokenSetId,
           mintedTimestamp: mintedTimestamp ?? null,
+          paymentTokens: { opensea: collection.paymentTokens } ?? null,
```
nofir, line 93: I think you're missing logic to update in here:
https://github.com/reservoirprotocol/indexer/blob/6b3bd6d5c6339a1fcf6483152e832e4531db0481/packages/indexer/src/models/collections/index.ts#L134

nofir (changes requested):

packages/indexer/src/jobs/token-updates/fetch-collection-metadata-job.ts
tv3636: not sure what you mean, something other than this? https://github.com/reservoirprotocol/indexer/pull/5044/files/1820d65df68cbf55c886e01a59cf8bcf952df8fc#diff-c98d5af98cf63c22904b691548fb3f095cdfd3afb1ee3573238e6b6472c076b1R78

packages/indexer/src/api/endpoints/collections/get-collection-supported-marketplaces/v2.ts
tv3636: yes was thinking we would run a backfill

packages/indexer/src/jobs/token-updates/fetch-collection-metadata-job.ts
nofir: `    const query = `
      UPDATE collections SET
        metadata = $/metadata:json/,
        name = $/name/,
        slug = $/slug/,
        token_count = $/tokenCount/,
        updated_at = now()
      WHERE id = $/id/
      RETURNING (
                  SELECT
                  json_build_object(
                    'name', collections.name,
                    'metadata', collections.metadata
                  )
                  FROM collections
                  WHERE collections.id = $/id/
                ) AS old_metadata
    `;

    const values = {
      id: collection.id,
      metadata: collection.metadata || {},
      name: collection.name,
      slug: collection.slug,
      tokenCount,
    };`


this is the query that updates the collection, so your payment tokens wont be included in the update, no?

tv3636: ah I see thanks, updated

nofir (approved):

PR #5084, opened 2023-06-13
feat: dynamic seaport conduits

lljxx1:


packages/indexer/src/migrations/1686645629521_seaport-conduit.sql
```diff
@@ -0,0 +1,19 @@
+-- Up Migration
+
+CREATE TABLE "seaport_conduits" (
```
georgeroman, line 3: Let's use the following schema:
```sql
CREATE TABLE "seaport_conduit_channels" (
  conduit_key BYTEA NOT NULL,
  channel BYTEA NOT NULL
);

ALTER TABLE "seaport_conduit_channels"
  ADD CONSTRAINT "seaport_conduit_channels_pk"
  PRIMARY KEY ("conduit_key", "channel");
```

Basically, the table should allow easy and efficient access to all open channels of a conduit.

packages/indexer/src/orderbook/orders/seaport-v1.5/index.ts
```diff
@@ -129,10 +129,13 @@ export const save = async (
           Sdk.SeaportBase.Addresses.SpaceIdConduitKey[config.chainId],
         ].includes(order.params.conduitKey)
       ) {
-        return results.push({
-          id,
-          status: "unsupported-conduit",
-        });
+        const conduits = await getConduits([order.params.conduitKey]);
```
georgeroman, line 132: Until everything is backfilled, let's leave this part of the code as it was before.

packages/indexer/src/sync/events/data/seaport.ts
```diff
@@ -76,3 +76,25 @@ export const orderValidated: EventData = {
     )`,
   ]),
 };
+
+export const newConduit: EventData = {
```
georgeroman, line 80: Thinking about it, a new conduit in itself is not relevant at all, so we should only be listening to the `channel-updated` events.

packages/indexer/src/sync/events/handlers/seaport.ts
```diff
@@ -446,6 +446,23 @@ export const handleEvents = async (events: EnhancedEvent[], onChainData: OnChain
...
+
+      case "seaport-new-conduit": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const conduit = parsedLog.args["conduit"].toLowerCase();
+        const conduitKey = parsedLog.args["conduitKey"].toLowerCase();
+        await saveConduit(conduit, `0x${conduitKey}`, baseEventParams.txHash);
+        break;
+      }
+
+      case "seaport-channel-updated": {
```
georgeroman, line 458: Whenever there is an updated channel, we should make an on-chain call to `getChannel` (on the conduit controller) to get all open channels and override in the database. This way we don't have to be careful about the ordering of the events since we'll always only fetch the latest available state.

image

georgeroman (changes requested):

PR #5103, opened 2023-06-13
fix: add error handling adding offsets

d3or:


packages/indexer/src/common/websocketPublisher.ts
```diff
@@ -22,8 +23,19 @@ export const addOffsetToSortedSet = async (
...
+  try {
+    if (!offset) {
+      return;
+    }
+    const stringOffset = String(offset);
+    const stringEvent = String(event.event);
+
+    await redisWebsocketPublisher.zadd(
+      "offsets",
+      `${stringOffset}-${stringEvent}`,
```
nofir, line 35: i think the timestamp needs to be before the value:
redis.zadd("sortedSet", 1, "one", 2, "dos", 4, "quatro", 3, "three");

nofir (changes requested):

nofir (approved):

nofir (approved):

PR #5143, opened 2023-06-15
feat: add cryptopunks module for aggregator 'router'

DessertHeart:
add cryptopunks module contract and its setup ts codes in router.ts


packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,145 @@
...
+        {} catch {
+            // Revert if specified
+            if (revertIfIncomplete) {
+                revert UnsuccessfulFill();
+            }
+        } 
+
+        // transfer the punk back to receiver
+        try 
+            PUNK_MARKETPLACE.transferPunk(
```
georgeroman, line 72: This will succeed in all cases, so we can simply move it inside the `{}` of the `try / catch` above.

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,145 @@
...
+                revert UnsuccessfulFill();
+            }
+        }
+    }
+
+    // --- Multiple sell Punks ---
+
+    function batchSellPunks(
+        ICryptoPunksMarket.SellOrder[] calldata sellOrders,
+        OfferParams calldata params
```
georgeroman, line 87: For simplicity and consistency with the rest of the modules, let's have this method fill a single offer rather than multiple ones.

georgeroman, line 87: Also, we should integrate fees:https://github.com/reservoirprotocol/indexer/blob/8cea061ab2dc27eb802e89e7b5da82cc83e04646/packages/contracts/contracts/router/modules/exchanges/LooksRareV2Module.sol#L114

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    function _sell(
+        ICryptoPunksMarket.SellOrder calldata sellOrder,
+        address seller,
+        bool revertIfIncomplete
+    ) internal {
+        // check if already have the punk from seller
+        require(
```
georgeroman, line 106: No need to have any assertions, the router will make sure the calldata is properly built

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,145 @@
...
+        {} catch {
+            // Revert if specified
+            if (revertIfIncomplete) {
+                revert UnsuccessfulFill();
+            }
+        }
+
+        // withdraw ETH
+        try
+            PUNK_MARKETPLACE.withdraw()
```
georgeroman, line 126: Let's move this inside the `try / catch` `{}` from above

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,145 @@
...
+        {} catch {
+            // Revert if specified
+            if (revertIfIncomplete) {
+                revert UnsuccessfulFill();
+            }
+        }
+
+        // transfer ETH back to seller
+        (bool success,) = payable(seller).call{value: sellOrder.price}("");
```
georgeroman, line 135: Let's use `_sendAllETH`:https://github.com/reservoirprotocol/indexer/blob/8cea061ab2dc27eb802e89e7b5da82cc83e04646/packages/contracts/contracts/router/modules/exchanges/BaseExchangeModule.sol#L156-L158

packages/sdk/src/router/v6/router.ts
```diff
@@ -825,6 +831,7 @@ export class Router {
     const nftxDetails: ListingDetails[] = [];
     const raribleDetails: ListingDetails[] = [];
     const superRareDetails: ListingDetails[] = [];
+    const cryptoPunkDetails: ListingDetails[] = [];
```
georgeroman, line 834: Now that we fill via the router, we can remove the direct filling logic:
image

packages/sdk/src/router/v6/router.ts
```diff
@@ -3779,6 +3842,40 @@ export class Router {
 
           break;
         }
+
+        //  Attention:
```
georgeroman, line 3846: I think splitting the approval and the sell transactions is risky since someone could front-run the sell transaction and send the punk to themselves instead of having it sold. For this reason, it probably make sense to disallow filling offers via the router, and always fill directly.

georgeroman (changes requested):

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
DessertHeart: I see, but I found that we also implemented multiple 'sell' in some modules like 'NFTXModule'. So in this case, we still should use 'single fill' instead?

packages/sdk/src/router/v6/router.ts
DessertHeart: totally agreed. So how about I just remove the logic for filling offers in router.ts?

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
DessertHeart: btw, for 'chrageFee' method in contract, is it a interface allowing aggregator(like us) charge extra tx fees? if it is, have u guys packed the method into APIs? if not, how could we do to make it?

packages/sdk/src/router/v6/router.ts
georgeroman: Yep, let's do that

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
georgeroman: If we remove the logic for filling offers via the router (due to the risks involved), this isn't needed anymore

packages/contracts/contracts/router/modules/exchanges/CryptoPunksModule.sol
```diff
@@ -0,0 +1,121 @@
...
+        } catch {
+            // Revert if specified
+            if (revertIfIncomplete) {
+                revert UnsuccessfulFill();
+            }
+        } 
+        
+    }
+
+    // --- Multiple sell Punks ---
```
georgeroman, line 74: As discussed, `batchSellPunks` and `_sell` aren't needed anymore.

georgeroman (changes requested):

PR #5159, opened 2023-06-15
Tv/platf 1110 collections websocket events postgres cdc kafka based

tv3636:


d3or (changes requested):
collection-websocket-events-trigger-queue needs to be added to allJobQueues in src/jobs/index.ts, view tokenWebsocketEventsTriggerQueue for example

d3or (approved):

PR #5207, opened 2023-06-19
feat: add excludeSources for asks and bids

tv3636:


ipeleg (commented):
Previous versions tags needs to be updated to x-deprecated

packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
```diff
@@ -0,0 +1,614 @@
...
+      }
+
+      if (query.excludeSources) {
+        const sources = await Sources.getInstance();
+
+        if (!Array.isArray(query.excludeSources)) {
+          query.excludeSources = [query.excludeSources];
+        }
+
+        (query as any).excludeSourceIds = query.excludeSources.map(
```
ipeleg, line 443: once sources are fetched let's filter any id 0 and if array is empty no need to apply the filter, relevant also to the other api and for the sources filter

ipeleg (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
tv3636: added a filter and check for empty sources list

ipeleg (approved):

PR #5211, opened 2023-06-20
feat: support payment processor

lljxx1:


packages/sdk/src/payment-processor/types.ts
```diff
@@ -0,0 +1,70 @@
...
+  listingNonce: BigNumberish;
+  offerNonce: BigNumberish;
+  listingMinPrice: BigNumberish;
+  offerPrice: BigNumberish;
+  listingExpiration: BigNumberish;
+  offerExpiration: BigNumberish;
+  tokenId: BigNumberish;
+  amount: BigNumberish;
+
+  sellerMasterNonce: BigNumberish;
```
georgeroman, line 31: Do we actually need these two? They don't seem to be present in the contract `MatchedOrder` struct.

packages/sdk/src/payment-processor/types.ts
```diff
@@ -0,0 +1,70 @@
...
+  buyerMasterNonce: BigNumberish;
+
+  listingSignature?: string;
```
georgeroman, line 34: Let's store the `v`, `r` and `s` directly instead of the compact signature, to match the contract code.

packages/sdk/src/payment-processor/types.ts
```diff
@@ -0,0 +1,70 @@
+import { BigNumberish } from "@ethersproject/bignumber";
+
+export enum TokenProtocols {
+  ERC721,
+  ERC1155,
+}
+
+export type MatchOrder = {
```
georgeroman, line 8: Let's rename this to `MatchedOrder`, to match the contract code.

packages/sdk/src/payment-processor/order.ts
```diff
@@ -0,0 +1,275 @@
...
+import { Contract } from "@ethersproject/contracts";
+import { AddressZero } from "@ethersproject/constants";
+import ExchangeAbi from "./abis/PaymentProcessor.json";
+import * as Addresses from "./addresses";
+import { TypedDataSigner } from "@ethersproject/abstract-signer";
+import * as Common from "../common";
+
+export class Order {
+  public chainId: number;
+  public params: Types.MatchOrder;
```
georgeroman, line 15: I think `MatchedOrder` should only be used when filling, otherwise we should store either `SaleApproval` or `OfferApproval` directly (since these are the actual order formats).

packages/sdk/src/payment-processor/types.ts
```diff
@@ -0,0 +1,70 @@
...
+  amount: BigNumberish;
+
+  sellerMasterNonce: BigNumberish;
+  buyerMasterNonce: BigNumberish;
+
+  listingSignature?: string;
+  offerSignature?: string;
+};
+
+export type SaleApproval = {
```
georgeroman, line 38: We should store the signatures in the `SaleApproval` and `OfferApproval` structs.

packages/sdk/src/payment-processor/order.ts
```diff
@@ -0,0 +1,275 @@
...
+    }
+  }
+
+  public async sign(signer: TypedDataSigner) {
+    const [types, value] = this.getEip712TypesAndValue();
+    const listingSignature = await signer._signTypedData(EIP712_DOMAIN(this.chainId), types, value);
+    this.params.listingSignature = listingSignature;
+  }
+
+  public async signOffer(signer: TypedDataSigner) {
```
georgeroman, line 32: We should have a single method for signing that detects the kind automatically (`sale-approval` vs `offer-approval` vs `collection-offer-approval`).

georgeroman (changes requested):

packages/indexer/src/sync/events/data/payment-processor.ts
```diff
@@ -0,0 +1,92 @@
...
+import { PaymentProcessor } from "@reservoir0x/sdk";
+
+import { config } from "@/config/index";
+import { EventData } from "@/events-sync/data";
+
+export const buySingleListing: EventData = {
+  kind: "payment-processor",
+  subKind: "payment-processor-buy-single-listing",
+  addresses: { [PaymentProcessor.Addresses.PaymentProcessor[config.chainId]?.toLowerCase()]: true },
+  topic: "0x7ed668b30822ae5c7db7b4a32f84e6645250bb6db0d1fc73eeb484c5b34b1d1b",
```
georgeroman, line 11: The first 3 events have the same topic, that seems wrong.

packages/indexer/src/sync/events/handlers/payment-processor.ts
```diff
@@ -0,0 +1,336 @@
...
+          nonce: orderNonce,
+          baseEventParams,
+        });
+        break;
+      }
+
+      case "payment-processor-master-nonce-invalidated": {
+        const parsedLog = eventData.abi.parseLog(log);
+        const maker = parsedLog.args["account"].toLowerCase();
+        const newNonce = parsedLog.args["none"].toString();
```
georgeroman, line 40: Typo, should be `nonce` instead of `none`

packages/sdk/src/payment-processor/addresses.ts
```diff
@@ -0,0 +1,5 @@
+import { ChainIdToAddress, Network } from "../utils";
+
+export const PaymentProcessor: ChainIdToAddress = {
+  [Network.Ethereum]: "0x009a1dc629242961c9e4f089b437afd394474cc0",
```
georgeroman, line 4: Let's also add the Goerli address in here, we'll be testing there.

packages/sdk/src/payment-processor/types.ts
```diff
@@ -0,0 +1,122 @@
...
+  listingNonce: string;
+  offerNonce: string;
+  listingMinPrice: string;
+  offerPrice: string;
+  listingExpiration: string;
+  offerExpiration: string;
+  tokenId: string;
+  amount: string;
+
+  sellerMasterNonce: string;
```
georgeroman, line 30: Is this and buyerMasterNonce needed for anything?

packages/sdk/src/payment-processor/exchange.ts
```diff
@@ -0,0 +1,112 @@
...
+  public contract: Contract;
+
+  constructor(chainId: number) {
+    this.chainId = chainId;
+    this.contract = new Contract(Addresses.PaymentProcessor[this.chainId], ExchangeAbi);
+  }
+
+  // --- Get nonce ---
+
+  public async getNonce(provider: Provider, user: string): Promise<BigNumber> {
```
georgeroman, line 22: Let's rename this to `getMasterNonce`

packages/indexer/src/sync/events/handlers/payment-processor.ts
```diff
@@ -0,0 +1,336 @@
...
+          maker,
+          minNonce: newNonce,
+          acrossAll: true,
+          baseEventParams,
+        });
+        break;
+      }
+
+      case "payment-processor-buy-single-listing": {
+        // const parsedLog = eventData.abi.parseLog(log);
```
georgeroman, line 54: Why do we need to use call tracing instead of the emitted events?

packages/indexer/src/orderbook/orders/payment-processor/index.ts
```diff
@@ -0,0 +1,403 @@
...
+      let source = await sources.getOrInsert("limitbreak.com");
+      if (metadata.source) {
+        source = await sources.getOrInsert(metadata.source);
+      }
+
+      // Handle: native Reservoir orders
+      const isReservoir = false;
+
+      // Handle: conduit
+      let conduit = Sdk.LooksRareV2.Addresses.Exchange[config.chainId];
```
georgeroman, line 287: This is wrong

packages/indexer/src/orderbook/orders/payment-processor/index.ts
```diff
@@ -0,0 +1,403 @@
...
+          Sdk.Common.Addresses.Eth[config.chainId],
+        ].includes(order.params.coin)
+      ) {
+        return results.push({
+          id,
+          status: "unsupported-payment-token",
+        });
+      }
+
+      // // Check: order is valid
```
georgeroman, line 97: We should probably enable this check

packages/indexer/src/orderbook/orders/payment-processor/index.ts
```diff
@@ -0,0 +1,403 @@
...
+
+export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
+  const results: SaveResult[] = [];
+  const orderValues: DbOrder[] = [];
+
+  const handleOrder = async ({ orderParams, metadata }: OrderInfo) => {
+    try {
+      const order = new Sdk.PaymentProcessor.Order(config.chainId, orderParams);
+      const id = order.hash();
+
```
georgeroman, line 43: Initially we only want to support orders on contracts with the default security policy id. So we should add a check for that as well (on-chain call is fine).

georgeroman (changes requested):

packages/sdk/src/payment-processor/types.ts
lljxx1: yeah, when validating the signature, both are needed

packages/indexer/src/sync/events/handlers/payment-processor.ts
lljxx1: need to compute the order ID via Matched Order

PR #5285, opened 2023-06-22
fix: looksrare goerli fee

jmaier6:
added other fee breakdown for goerli

### DIFF
--- a/packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
+++ b/packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
@@ -189,7 +189,7 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
               {
                 kind: "marketplace",
                 recipient: "0xdbBE0859791E44B52B98FcCA341DFb7577C0B077",
-                bps: 50,
+                bps: 200,
               },
             ];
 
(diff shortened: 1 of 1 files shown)


packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -176,13 +176,22 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
...
+              {
+                kind: "marketplace",
+                recipient: "0x1838de7d4e4e42c8eb7b204a91e28e9fad14f536",
+                bps: 50,
+              },
+            ]
+          : [
+              {
+                kind: "marketplace",
+                recipient: "0xdbBE0859791E44B52B98FcCA341DFb7577C0B077",
```
georgeroman, line 191: Let's lowercase the address

packages/indexer/src/orderbook/orders/looks-rare-v2/index.ts
```diff
@@ -176,13 +176,22 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
...
+                bps: 50,
```
georgeroman, line 192: Shouldn't this be 200?

georgeroman (changes requested):

PR #5318, opened 2023-06-27
feat: compare token set new top bid to collection top bid for triggering websocket top-bid.changed

d3or:


packages/indexer/src/jobs/token-set-updates/utils.ts
```diff
@@ -43,6 +44,12 @@ export async function processTopBid(payload: topBidPayload, queueName: string) {
                     ORDER BY orders.value DESC
                     LIMIT 1
                   ) y ON TRUE
+                  LEFT JOIN LATERAL (
```
nofir, line 47: the collection id is NOT set on single token tokenset but maybe you can parse the token set id and get the collection based on the token contract and id (similar to how we do the criteria for orders)

packages/indexer/src/jobs/token-set-updates/top-bid-queue.ts
```diff
@@ -70,7 +70,8 @@ export const jobProcessor = async (job: Job) => {
                 WITH x AS (
```
nofir, line 70: I dont think we need to make changes in this file as we moved into rabbitmq. we should probably remove :
top-bid-queue.ts
top-bid-single-token-queue.ts
(and all assoicated references)

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/top-bid-websocket-events-trigger-queue.ts
```diff
@@ -95,6 +95,18 @@ if (config.doBackgroundWork && config.doWebsocketServerWork && config.kafkaBroke
...
 
+        const topBidOnCollection = await redis.get(`top-bid:${order.collection_id}`);
+        if (topBidOnCollection && Number(topBidOnCollection) > Number(order.price)) {
+          logger.warn(
+            QUEUE_NAME,
+            `Top bid on collection is higher than current bid. data=${JSON.stringify(data)}`
+          );
+
+          return;
+        } else {
```
nofir, line 106: that is wrong, the top bid on the token is not necessarily the top bid on the collection is its a trait / single token offer.

packages/indexer/src/jobs/websocket-events/top-bid-websocket-events-trigger-queue.ts
```diff
@@ -95,6 +95,18 @@ if (config.doBackgroundWork && config.doWebsocketServerWork && config.kafkaBroke
           return;
         }
 
+        const topBidOnCollection = await redis.get(`top-bid:${order.collection_id}`);
+        if (topBidOnCollection && Number(topBidOnCollection) > Number(order.price)) {
```
nofir, line 99: 1. Why not use what we [already cache](https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/models/top-bids-caching/index.ts#L11)? 
2. i think the check needs to happen only if its a trait / single token bid (maybe not an issue if we assume that the collection offer will actually be higher). 
3. you should be comparing the order value (rather than the price).
3. Wdyt on passing the "orderValue" in the payload to the job? so we can skip the query to the db?

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/top-bid-websocket-events-trigger-queue.ts
```diff
@@ -32,6 +32,17 @@ if (config.doBackgroundWork && config.doWebsocketServerWork && config.kafkaBroke
     async (job: Job) => {
       const { data } = job.data as EventInfo;
 
+      // if its a single token top bid, check if its lower than the top bid on the collection and skip if it is
+      const topBidOnCollection = await redis.get(`collection-top-bid:${data.collectionId}`);
+      if (topBidOnCollection && Number(topBidOnCollection) > Number(data.orderValue)) {
```
nofir, line 37: What would happen in the case of a new collection offer that is the new top bid? because we set the cache when we calculate the new collection top bid, it wont fire the ws (coz its not higher than the cache, its equal), right? i would maybe pass a skipCollectionTopBidCheck or the bid criteria into the event and only do it for trait/ single token bids

nofir (changes requested):

nofir (approved):

packages/indexer/src/jobs/token-set-updates/utils.ts
```diff
@@ -100,11 +100,11 @@ export async function processTopBid(payload: topBidPayload, queueName: string) {
         tokenSetTopBid[0].topBuyId &&
         _.isNull(tokenSetTopBid[0].collectionId)
       ) {
-        //  Only trigger websocket event for non collection offers.
         await WebsocketEventRouter({
           eventKind: WebsocketEventKind.NewTopBid,
           eventInfo: {
             orderId: tokenSetTopBid[0].topBuyId,
+            isSingleTokenBid: true,
```
nofir, line 107: small fix, this can also be trait bid so maybe the flag should be: isNonCollectionBids / validateCollectionTopBid, etc

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/top-bid-websocket-events-trigger-job.ts
```diff
@@ -90,6 +92,19 @@ export class TopBidWebSocketEventsTriggerJob extends AbstractRabbitMqJobHandler
         return;
       }
 
+      if (data?.isSingleTokenBid) {
+        if (Number(order.value) < Number(order.top_buy_value)) {
+          logger.warn(
```
nofir, line 97: we will need to watch over this log to make sure we remove it after monitoring for a bit on prod, could result in a lot of logs

nofir (approved):

PR #5326, opened 2023-06-27
feat: add originated_at for asks and bids

tv3636:


packages/indexer/src/api/endpoints/orders/get-orders-bids/v5.ts
```diff
@@ -312,6 +313,7 @@ export const getOrdersBidsV5Options: RouteOptions = {
           orders.is_reservoir,
           extract(epoch from orders.created_at) AS created_at,
           extract(epoch from orders.updated_at) AS updated_at,
+          extract(epoch from orders.originated_at) AS originated_at,
```
nofir, line 316: you dont need the extract, we do that for fields we sort by as far as i remember and then you can change line 688 to now do *1000

nofir (approved):

PR #5327, opened 2023-06-27
Add ArtBlocks Engine contract for Prohibition on Arbitrum

tlip:


packages/indexer/src/config/network.ts
```diff
@@ -623,6 +623,14 @@ export const getNetworkSettings = (): NetworkSettings => {
         lastBlockLatency: 5,
         headBlockDelay: 10,
         subDomain: "api-arbitrum",
+        washTradingExcludedContracts: [
+          // Prohibition Contracts - ArtBlocks Engine
+          "0x47A91457a3a1f700097199Fd63c039c4784384aB",
```
nofir, line 628: Needs to be lowecase (also the one below)

nofir (changes requested):

nofir (approved):

PR #5332, opened 2023-06-28
$Sheesh token

dturska12:
Add support for Sheesh token


packages/indexer/src/config/network.ts
```diff
@@ -310,29 +310,14 @@ export const getNetworkSettings = (): NetworkSettings => {
             },
           ],
           [
-            "0xda9f05a3e133c2907e7173495022a936a3808d45",
```
nofir, line 313: Why were Nelkcoin and RugBank Token removed?

dturska12: When submitting the pr it said they were causing a conflict, can resubmit if need be. I must have must read what the issue was

nofir: yeah, just seems like a mistake during merge

dturska12: fixed and sumbitted, sorry about that last one

PR #5333, opened 2023-06-28
feat: support erc1155 for x2y2

lljxx1:


packages/indexer/src/orderbook/orders/common/helpers.ts
```diff
@@ -218,17 +216,18 @@ export const isListingOffChainCancelled = async (
...
   if (["blur", "x2y2"].includes(orderKind)) {
     // Blur and X2Y2 orders can get off-chain cancelled if the user
     // transferred to another wallet or revoked the approval
-    let conduit: string;
-    if (orderKind === "blur") {
-      conduit = Sdk.Blur.Addresses.ExecutionDelegate[config.chainId];
-    } else {
-      conduit = Sdk.X2Y2.Addresses.Erc721Delegate[config.chainId];
-    }
+    // let conduit: string;
```
georgeroman, line 225: This shouldn't be commented out

georgeroman (changes requested):
Two things that are missing:
- see screenshot below, we'll need a different encoding for ERC1155 orders (I think the amount is also encoded in there, but you should find an example and confirm that)
- once we add support for an amount, we should have an additional check in the order saving logic is make sure that ERC1155 listings and bids have a quantity of 1 (reject all the other ones)

packages/indexer/src/orderbook/orders/common/helpers.ts
lljxx1: I moved them to the params

PR #5356, opened 2023-06-29
fix: 401 error on server setup

compojoom:
In docker-compose.yaml the rabbitMQ container is started with a username and password. However when we are setuping our server we are trying to connect to rabbitMQ over http without basic auth. This ends up in a 401 error, that we don't catch and the initialization fails.


ipeleg:
You can set the RABBIT_URL to be guest:guest@127.0.0.1 which will take care of the authentication

PR #5368, opened 2023-06-29
update token image url

dturska12:
update token image for sheesh

### DIFF
--- a/packages/indexer/src/elasticsearch/indexes/activities/index.ts
+++ b/packages/indexer/src/elasticsearch/indexes/activities/index.ts
@@ -1490,6 +1490,160 @@ export const updateActivitiesCollectionMetadata = async (
   return keepGoing;
 };
 
+export const updateActivitiesCollectionMetadataV2 = async (
+  collectionId: string,
+  collectionData: { name: string | null; image: string | null }
+): Promise<boolean> => {
+  let keepGoing = false;
+
+  const should: any[] = [
+    {
+      bool: collectionData.name
+        ? {
+            must_not: [
+              {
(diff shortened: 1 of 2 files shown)


packages/indexer/src/config/network.ts
```diff
@@ -335,6 +335,19 @@ export const getNetworkSettings = (): NetworkSettings => {
...
+          [
+            "0xbb4f3ad7a2cf75d8effc4f6d7bd21d95f06165ca",
+            {
+              contract: "0xbb4f3ad7a2cf75d8effc4f6d7bd21d95f06165ca",
+              name: "Sheesh",
+              symbol: "SHS",
+              decimals: 18,
+              metadata: {
+                image:
+                  "https://bafybeic2ukraukxbvs7mn5f5xqnqkr42r5exxjpij4fmw4otiows2zjzbi.ipfs-public.thirdwebcdn.com/Screenshot_2023-06-15_003656.png",
```
nofir, line 347: The original pr was already merged and has the entry for Sheesh, are you sure you have the latest version?

nofir: You should only need to change the url

nofir (changes requested):

packages/indexer/src/config/network.ts
nofir: the entry (contract address) should be lowercased. You also have merge conflicts which you need to resolve. I think you need to make sure you have the latest version of main on your end. github would not detect it has a new addition just because of the case (it would show as a change in the existing code)

PR #5400, opened 2023-07-03
feat: realtime use v2 only, no more v1

d3or:


nofir (approved):
if it looks stable, lets just cleanup the unused code after

PR #5403, opened 2023-07-03
feat: add base mainnet

tv3636:


nofir (dismissed):
looks good. if opensea supports base, you also need to add:
https://github.com/reservoirprotocol/indexer/blob/0de7307ccdd3d12c03621569fe92a28028025795/packages/indexer/src/websockets/opensea/utils.ts#L5

nofir (approved):

PR #5414, opened 2023-07-04
fix: don't write to elastic if not configured

compojoom:
The write to elastic search was failing, since I have no elastic search configured.


packages/indexer/src/jobs/activities/save-pending-activities-job.ts
```diff
@@ -55,6 +55,9 @@ export class SavePendingActivitiesJob extends AbstractRabbitMqJobHandler {
   }
 
   public async addToQueue() {
+    if (!config.doElasticsearchWork) {
```
nofir, line 58: Makes sense. Can you also add it to the cron schedule (line #71)?

nofir (approved):

PR #5432, opened 2023-07-06
deor/platf 1619 enable changed in tokens ws

d3or:
- feat: enable changed on token ws
- feat: enable tokens changed on polygon in prod


nofir (dismissed):

packages/indexer/src/migrations/1689087567231_alter_replication_full.sql
```diff
@@ -4,6 +4,7 @@ ALTER TABLE tokens REPLICA IDENTITY FULL;
 ALTER TABLE orders REPLICA IDENTITY FULL;
 ALTER TABLE fill_events_2 REPLICA IDENTITY FULL;
 ALTER TABLE collections REPLICA IDENTITY FULL;
+ALTER TABLE nft_transfer_events REPLICA IDENTITY FULL;
```
nofir, line 7: didnt we say that each story for adding the change should do the migration for the dedicated table? not really an issue if we're going to do that anyway.

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
```diff
@@ -228,31 +228,31 @@ if (config.doBackgroundWork && config.doWebsocketServerWork && config.kafkaBroke
...
+            for (const key in changedMapping) {
+              // eslint-disable-next-line
+              // @ts-ignore
+              if (data.before[key] !== data.after[key]) {
+                changed.push(key);
+              }
+            }
+
+            if (!changed.length) {
+              return;
```
nofir, line 245: can we add a log here to see how many / which changed dont get propogated?  just to make sure we didnt miss anything. we'll remove at some point basically have the before and after

nofir (approved):

nofir (approved):

PR #5439, opened 2023-07-06
feat: sale parse for blur v2

lljxx1:


packages/indexer/src/sync/events/handlers/blur-v2.ts
```diff
@@ -0,0 +1,161 @@
...
+            selector: "0x3925c3c3",
+            name: "takeAsk",
+          },
+          {
+            selector: "0x70bce2d6",
+            name: "takeAskSingle",
+          },
+          // {
+          //   selector: "0x336d8206",
+          //   name: "takeAskSinglePool",
```
georgeroman, line 42: Should we support this and `takeAskPool`?

PR #5442, opened 2023-07-06
fix: db crashes when used with a connection pool

compojoom:
The database we use is behind a pgbouncer setup. PGbouncer doesn’t support statement_timeout and when we issue queries with it the queries are failing with: Unhandled rejection: error: unsupported startup parameter: statement_timeout

If one uses a connection pool/pgbouncer it’s not necessary to use statement_timeout as the pool takes care of the connection.


packages/indexer/src/config/index.ts
```diff
@@ -33,6 +33,7 @@ export const config = {
   disableRealtimeMetadataRefresh: Boolean(Number(process.env.DISABLE_REALTIME_METADATA_REFRESH)),
 
   databaseUrl: String(process.env.DATABASE_URL),
+  databaseUsesConnectionPool: Boolean(Number(process.env.DATABASE_USES_CONNECTION_POOL)),
```
ipeleg, line 36: let's change this to
`dbStatementTimeout: process.env.DB_STATEMENT_TIMEOUT ? Number(process.env.STATEMENT_TIMEOUT) : undefined,`

packages/indexer/src/common/db.ts
```diff
@@ -65,7 +65,7 @@ export const ridb = pgp({
   max: config.chainId === 1 ? 250 : 60,
   connectionTimeoutMillis: 30 * 1000,
   query_timeout: 5 * 60 * 1000,
-  statement_timeout: 5 * 60 * 1000,
+  statement_timeout: config.databaseUsesConnectionPool ? undefined : 5 * 60 * 1000,
```
ipeleg, line 68: let's change this to `statement_timeout: config. dbStatementTimeout ?? 5 * 60 * 1000,` and all other occurences

ipeleg (changes requested):

packages/indexer/src/common/db.ts
compojoom: But all of the `statement_timeouts` are actually different? Do you really want to enforce the same value set in the
.env file for all queries?

ipeleg: that's a valid point, let's instead just change the name `databaseUsesConnectionPool` to `disableDatabaseTimeout`, i feel the name `databaseUsesConnectionPool` not properly representing what will happen if setting it, unless you dig in the code

compojoom: I changed it to disableDatabaseStatementTimeout and the env var is: DATABASE_DISABLE_STATEMENT_TIMEOUT
and if people know about statement_timeout in postgres they should kind of understand what this is about.

ipeleg (approved):

PR #5478, opened 2023-07-07
fix: chainId added to top-bid-oracle response

jmaier6:


jmaier6:
@georgeroman I noticed the top-bid-oracle was 500 missing chainId in response.

PR #5511, opened 2023-07-10
feat: reenable originatedAt for x2y2

tv3636:


packages/indexer/src/orderbook/orders/x2y2/index.ts
```diff
@@ -314,7 +314,7 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
         missing_royalties: missingRoyalties,
         normalized_value: normalizedValue,
         currency_normalized_value: normalizedValue,
-        // originated_at: metadata.originatedAt ? new Date(metadata.originatedAt) : null,
+        originated_at: metadata.originatedAt ? new Date(metadata.originatedAt).toISOString() : null,
```
nofir, line 317: what is the format is comes from the relayer?

nofir: we can but it will change on the relayer only if we change it as it does not depend on the x2y2 response

tv3636: ok simplified it a bit

nofir (approved):

PR #5516, opened 2023-07-11
add taker parameter for /execute/bid/v5 API

Anorth1997:


georgeroman:
@Anorth1997 Simply passing in the `taker` field to the build params won't result in the orders being private. You'll need to make sure the SDK builders actually support the `taker` parameter and build private orders. Also, note that some exchanges don't support private orders (eg. Blur). Let me know if you're interested in continuing this PR, otherwise will close.

PR #5532, opened 2023-07-12
feat: disable approval ws

d3or:


nofir (approved):
you also need to move it from debezium, right?

PR #5537, opened 2023-07-12
feat: sync events by address and fetch all logs

compojoom:
Generally when filtering by and address we want to reduce the amount of events we process/store in the database. However most of the time we are not only interested in the events only emitted by a particular contract, but also in the other events in the transaction. For example, if we want to fetch all the events for particular NFT I would argue that a sale on opensea is also event related to the NFT, but since the orderFullfilled event from seaport is not emitted by the contract address we filter on, we end up ommitting the events. I’ve added a fetchAllTxLogs param to the syncDetail when filtering by address. This will basically perform additional eth_getTransactionReceipt calls for every transactionHash. In some situatioons those calls might be redundant, but most of the time if we are dealing with sales on marketplaces we want the additional events.

### DIFF
--- a/packages/indexer/src/sync/events/index.ts
+++ b/packages/indexer/src/sync/events/index.ts
@@ -1,4 +1,4 @@
-import { Filter } from "@ethersproject/abstract-provider";
+import { Filter, Log } from "@ethersproject/abstract-provider";
 import _, { now } from "lodash";
 import pLimit from "p-limit";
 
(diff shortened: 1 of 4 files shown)


packages/indexer/src/sync/events/index.ts
```diff
@@ -350,137 +358,160 @@ export const syncEvents = async (
...
-            subKind: eventData.subKind,
-            baseEventParams,
-            log,
-          });
+      let allLogs: Log[] = [];
+      const uniqueTxHashes = [...new Set(logs.map((log) => log.transactionHash))];
+
+      for (const txHash of uniqueTxHashes) {
+        try {
+          const txReceipt = await baseProvider.getTransactionReceipt(txHash);
```
georgeroman, line 373: We have a utils method `fetchTransactionLogs` or something similar which also saves everything in the database.

georgeroman:
Closing this for now

PR #5540, opened 2023-07-12
Integrate Midaswap

LXVC:


packages/indexer/src/sync/events/data/midaswap.ts
```diff
@@ -0,0 +1,94 @@
...
+        uint256 binAmount
+    )`,
+  ]),
+};
+
+export const positionBurned: EventData = {
+  kind: "midaswap",
+  subKind: "midaswap-position-burned",
+  topic: "0x399008f8cf2698b523adce30291bfe4efa73018ed7095d73895bbf77d6e2ed2c",
+  numTopics: 4,
```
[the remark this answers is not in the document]
LXVC: correct!

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1168 @@
...
+          );
+
+          if (orderResult.length) {
+            // remove orders by pool && lp token id
+            const ids = orderResult
+              .filter((item) => item.raw_data.lpTokenId === lpTokenId)
+              .map((item) => item.id);
+            if (ids.length) {
+              await idb.none(`DELETE FROM orders WHERE orders.id IN ($/ids:list/)`, {
+                ids,
```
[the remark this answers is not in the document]
LXVC: Will this automatically set the 'fillability_status' field of these orders to 'cancelled'?

packages/indexer/src/orderbook/orders/index.ts
```diff
@@ -723,6 +733,15 @@ export const generateBidDetailsV6 = async (
...
 
+    case "midaswap": {
+      const sdkOrder = new Sdk.Midaswap.Order(config.chainId, order.rawData);
+      return {
+        kind: "midaswap",
+        ...common,
+        order: sdkOrder,
+      };
+    }
+
```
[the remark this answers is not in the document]
LXVC: Other exchanges also have two of the same logic in this file

packages/indexer/src/orderbook/orders/midaswap/index.ts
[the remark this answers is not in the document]
lljxx1: yeah. link them by the `ordeId`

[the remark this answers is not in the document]
lljxx1: let's only delete the orders that are not `filled`.

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1118 @@
...
+        }
+        case "midaswap-buy-erc721": {
+          if (_.isUndefined(tradeBin) || _.isUndefined(lpTokenId) || _.isUndefined(nftId)) {
+            return;
+          }
+
+          const id = getSellOrderId(orderParams.pool, nftId, lpTokenId);
+
+          // delete sell order
+          await idb.none(`DELETE FROM orders WHERE orders.id = $/id/`, { id });
```
lljxx1, line 477: If the order may be `filled`, please avoid deleting the order directly

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1118 @@
...
+            ]);
+
+            if (!tokenSetId) {
+              throw new Error("No token set available");
+            }
+
+            // Handle: source
+            const sources = await Sources.getInstance();
+            const source = await sources.getOrInsert("midaswap");
+
```
lljxx1, line 586: same as above

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1118 @@
...
+                  tokenId: nftId,
+                },
+              ]);
+              if (!tokenSetId) {
+                throw new Error("No token set available");
+              }
+
+              // Handle: source
+              const sources = await Sources.getInstance();
+              const source = await sources.getOrInsert("midaswap");
```
lljxx1, line 920: same as above

packages/indexer/src/orderbook/orders/midaswap/index.ts
LXVC: One question, if an order fires the 'filled' event in handler, will the change happen before or after the 'save' function is executed? Or not sure?

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1118 @@
...
+            }
+
+            // Handle: source
+            const sources = await Sources.getInstance();
+            const source = await sources.getOrInsert("midaswap");
+
+            const validFrom = `date_trunc('seconds', to_timestamp(${orderParams.txTimestamp}))`;
+            const validTo = `'Infinity'`;
+            tmpPriceList.forEach(() => {
+              const id = getBuyOrderId(pool.address, lpTokenId, randomUUID());
```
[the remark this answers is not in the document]
LXVC: There is a solution, but it will cause some problems. The parameter for `getBuyOrderId` will become `(pool: string, lpTokenId: string, bin: string, index: number)`. This is not a problem when creating a `buy` order, but when the order is traded, there is no `index`. Therefore, when trading, I have to search for the `index` from 0 until I find a "fillable" order. Is this acceptable?

packages/indexer/src/sync/events/handlers/index.ts
```diff
@@ -61,7 +62,11 @@ export type EventsBatch = {
 // Map each high-level event kind to its corresponding handler
 export const eventKindToHandler = new Map<
   EventKind,
-  (e: EnhancedEvent[], d: OnChainData, backfill?: boolean) => Promise<void>
+  (
+    e: EnhancedEvent[],
+    d: OnChainData,
+    backfill?: boolean
+  ) => Promise<void> | Promise<OnChainData | undefined>
```
georgeroman, line 69: Why this change? Seems unrelated to the integration.

packages/sdk/src/midaswap/order.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public static binToPriceFixed = (bin: number, decimal = 18, toFixedNumber = 18) => {
+    const powValue = bin - 8388608;
+    const b = new Decimal(10).pow(18 - decimal);
+    const price = new Decimal(1.0001).pow(powValue).times(b).toFixed(toFixedNumber);
+    return String(price);
+  };
+
+  public static getSellPrice = (bin: number, freeRate = 0, roralty = 0) => {
```
georgeroman, line 27: Should be `royalty`, below as well.

packages/contracts/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@tenderly/hardhat-tenderly": "^1.6.1",
     "@uniswap/permit2-sdk": "^1.2.0",
+    "decimal.js": "^10.4.3",
```
georgeroman, line 15: Where is this used? If possible, let's not add any new dependencies.

packages/indexer/src/common/joi.ts
```diff
@@ -394,6 +396,27 @@ export const getJoiDynamicPricingObject = async (
         ),
       },
     };
+  } else if (kind === "midaswap") {
```
georgeroman, line 399: Can we just add a new `pool` field in the raw data which includes the LP id? Would try to avoid adding another branch.

packages/indexer/src/orderbook/orders/utils.ts
```diff
@@ -16,6 +16,14 @@ export type OrderMetadata = {
   target?: string;
   originatedAt?: string;
   fromOnChain?: boolean;
+
+  eventName?: string;
```
georgeroman, line 20: These fields shouldn't be here, this is a generic type that covers all orders so we should avoid adding anything specific to a particular exchange.

georgeroman: If needed, should include these fields in the `orderParams`.

packages/sdk/src/midaswap/order.ts
```diff
@@ -0,0 +1,69 @@
+import * as Types from "./types";
+import { lc, s } from "../utils";
+import Decimal from "decimal.js";
```
georgeroman, line 3: As mentioned in one comment above, let's try to avoid adding a new dependency.

georgeroman (changes requested):

packages/indexer/src/orderbook/orders/midaswap/index.ts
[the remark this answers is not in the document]
georgeroman: Could we use call tracing to get the index from any sub-calls? See how we handle sudoswap-v2 sales.

packages/indexer/src/orderbook/orders/midaswap/index.ts
georgeroman: The save will happen after.

packages/sdk/src/midaswap/order.ts
LXVC: Sorry, I don't understand. Should I rename `getSellPrice`?

packages/indexer/src/common/joi.ts
LXVC: So the `pool` field will not be an address, will be `${pair}_${lpTokenId}`, is that OK?

packages/indexer/src/orderbook/orders/midaswap/index.ts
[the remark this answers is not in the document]
LXVC: No can do. Our contract does not store index information, the NFTs under the same lpTokenId are unordered from the contract's perspective, index is an off-chain concept.

packages/sdk/src/midaswap/order.ts
LXVC: I'll try.

packages/indexer/src/common/joi.ts
georgeroman: Yes, just make sure to keep the `pair` field, which is gong to be an address.

georgeroman:
Closing this, already merged in a different PR

PR #5563, opened 2023-07-12
Tv/platf 1673 support token attribute ws

tv3636:


packages/indexer/src/jobs/websocket-events/attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,179 @@
...
+                    'value', ta.value,
+                    'createdAt', ta.created_at,
+                    'tokenCount', attributes.token_count,
+                    'onSaleCount', attributes.on_sale_count,
+                    'floorAskPrice', attributes.floor_sell_value::TEXT,
+                    'topBidValue', attributes.top_buy_value::TEXT
+                  )
+                )
+            FROM token_attributes ta
+            JOIN attributes
```
nofir, line 60: is there any impact as query performance if joining with attributes? we might be able to live without it

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,171 @@
...
+    },
+    removeOnComplete: 1000,
+    removeOnFail: 1000,
+    timeout: 60000,
+  },
+});
+new QueueScheduler(QUEUE_NAME, { connection: redis.duplicate() });
+
+const changedMapping = {
+  contract: "token_attributes.contract",
```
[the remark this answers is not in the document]
nofir: we decided to check if updated is even a scenario

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,171 @@
...
+import { config } from "@/config/index";
+
+import { randomUUID } from "crypto";
+import _ from "lodash";
+
+import { publishWebsocketEvent } from "@/common/websocketPublisher";
+import { redb } from "@/common/db";
+import { fromBuffer, toBuffer } from "@/common/utils";
+
+const QUEUE_NAME = "token-attribute-websocket-events-trigger-queue";
```
nofir, line 14: needs to change to a rabbitmq job

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,157 @@
+import { logger } from "@/common/logger";
```
nofir, line 1: can change the name to -job instead of -queue? that has been the approach so far.

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export type TokenAttributeWebsocketEventsTriggerQueueJobPayload = {
+  data: TokenAttributeWebsocketEventInfo;
+};
+
+const changedMapping = {
+  collection_id: "collection.id",
+  key: "key",
+  value: "value",
+  updated_at: "updatedAt",
```
nofir, line 16: i dont think "updated_at" should be in the changed mapping. if the only thing that change is the updated_at, then we should not send an event

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-queue.ts
```diff
@@ -0,0 +1,157 @@
...
+                  changed.push(key);
+                }
+              }
+
+              if (!changed.length) {
+                return;
+              }
+              */
+
+            logger.info(this.queueName, "update token attributes");
```
nofir, line 95: please add the before / after so we can understand better what are the usecases

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,154 @@
...
+  backoff = {
+    type: "exponential",
+    delay: 1000,
+  } as BackoffStrategy;
+
+  protected async process(payload: TokenAttributeWebsocketEventsTriggerQueueJobPayload) {
+    const { data } = payload;
+
+    try {
+      const baseQuery = `
```
nofir, line 29: i realized that this needs to be changed to use the data from the data.after as we have all the data alredy (similar to the stories to switch to using cdc data)

tv3636: isn't it already? line 47 and 48

nofir: i meant that there is no need to make a query to the db which you are doing in lines 29 - 45

packages/indexer/src/jobs/websocket-events/token-attribute-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      await publishWebsocketEvent({
+        event: eventType,
+        tags: {
+          token_id: data.after.token_id,
+          contract: data.after.contract,
+        },
+        changed,
+        data: data.after,
```
nofir, line 72: i see you changed to use the data from the event but you should still keep the same response standard:
`{    
            token: {
                  contract: fromBuffer(r.contract),
                  tokenId: r.token_id,
                },
                collection: {
                  id: r.collection_id,
                },
                key: r.key,
                value: r.value,
                createdAt: r.created_at,
                updatedAt: r.updated_at,
              }`

nofir (changes requested):

nofir (approved):

nofir (approved):

PR #5586, opened 2023-07-13
feat: add volume and pricing to recent sales

lunarmayor:


packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -188,21 +188,25 @@ export const getChainStatsFromActivity = async () => {
...
-                  range: {
-                    timestamp: {
-                      gte: period.startTime,
+            constant_score: {
+              filter: {
+                bool: {
+                  filter: [
+                    {
+                      terms: {
+                        type: ["sale", "mint"],
```
nofir, line 197: I see you are also querying mints, i did not backfill the pricingDecimal for mints which i guess makes sense for paid mints. so i might need to do that before you make the change.

also, to filter out free mints, you might want to add a filter to only include price > 0

nofir (changes requested):

packages/indexer/src/elasticsearch/indexes/activities/index.ts
lunarmayor: I'm also aggregating the count so still need to include the free mints

nofir (approved):

PR #5603, opened 2023-07-15
Tv/platf 1621 enable changed in sales ws

tv3636:


packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      let eventType = "";
+      const changed = [];
+      if (data.trigger === "insert") eventType = "sale.created";
+      else if (data.trigger === "update") {
+        // if isDeleted is true, then it's a delete event
+        if (r.is_deleted) eventType = "sale.deleted";
+        else {
+          eventType = "sale.updated";
+          if (data.before) {
```
nofir, line 159: seems like we are checking the changed logic also for deleted scenario which means that the deleted event wont get triggered

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,227 @@
...
+            for (const key in changedMapping) {
+              // eslint-disable-next-line
+              // @ts-ignore
+              if (data.before[key] !== data.after[key]) {
+                changed.push(key);
+              }
+            }
+
+            if (!changed.length) {
+              return;
```
nofir, line 169: i think a general approach for the prs you are making for "changed" is to add a log here so whenever the update was triggered for a field that is not in the changed, we will log it. (same for transfers, token attributes, etc) , it will also work for when the changed is empty.

nofir (changes requested):

nofir (dismissed):

nofir (approved):

nofir (approved):

nofir (approved):

PR #5611, opened 2023-07-17
feat: enabled changed on transfers

tv3636:


packages/indexer/src/jobs/index.ts
```diff
@@ -29,7 +29,6 @@ import * as backfillRefreshCollectionMetadata from "@/jobs/backfill/backfill-ref
 
 import * as askWebsocketEventsTriggerQueue from "@/jobs/websocket-events/ask-websocket-events-trigger-queue";
 import * as bidWebsocketEventsTriggerQueue from "@/jobs/websocket-events/bid-websocket-events-trigger-queue";
-import * as transferWebsocketEventsTriggerQueue from "@/jobs/websocket-events/transfer-websocket-events-trigger-queue";
```
nofir, line 32: I would make another pr to remove the queue, otherwise, you would lose visibility to when the queue is empty of any pending messages.

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
```diff
@@ -1,110 +1,113 @@
-import { Job, Queue, QueueScheduler, Worker } from "bullmq";
```
nofir, line 1: to continue the comment from before, you should have created a new file with the rabbitmq job. im sorry if i wasnt clear, it mainly to not lose any messages in the queue

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
```diff
@@ -1,110 +1,113 @@
...
-          amount: data.amount,
-          block: data.block,
-          txHash: data.tx_hash,
-          logIndex: data.log_index,
-          batchIndex: data.batch_index,
-          timestamp: data.timestamp,
-        };
+      let eventType = "";
+      //const changed = [];
+      if (data.after.is_deleted) eventType = "transfer.deleted";
```
nofir, line 48: it might make sense to check if data.before.is_deleted === 0 also so we wont trigger a deleted more than once. @d3or wdyt?

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-queue.ts
```diff
@@ -1,110 +1,113 @@
...
-        });
-      } catch (error) {
-        logger.error(
-          QUEUE_NAME,
-          `Error processing websocket event. data=${JSON.stringify(data)}, error=${JSON.stringify(
-            error
-          )}`
-        );
-        throw error;
+          logger.info(this.queueName, "update transfer");
```
nofir, line 67: please add the after / before to the log so we can understand what has been updated. I would either turn it into an error log or make sure you create an alarm for the info log. otherwise, it might be hard to track if its actually happening

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,140 @@
...
+        timestamp: data.after.timestamp,
+      };
+
+      let eventType = "";
+      //const changed = [];
+      if (data.after.is_deleted) eventType = "transfer.deleted";
+      else if (data.trigger === "insert") eventType = "transfer.created";
+      else if (data.trigger === "update") {
+        eventType = "transfer.updated";
+        if (data.before) {
```
nofir, line 52: why do we need the if (data.before)?

nofir (approved):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
nofir: Also, maybe change the logic to only do it on update, here it will also trigger the log for when a transfer is deleted.

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
tv3636: not sure about `if (data.before)` to be honest, copied that from [the PR to enable changed on tokens](https://github.com/reservoirprotocol/indexer/pull/5432/files#diff-a6beb83a9d20261ef8a99b15b515c346fffdf46b153465ce30dfaa0e6c6cda93R235)

nofir (approved):

d3or (approved):

nofir (dismissed):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
```diff
@@ -0,0 +1,132 @@
...
+            }
+          }
+
+          if (!changed.length) {
+            logger.info(
+              this.queueName,
+              `No changes detected for event. before=${JSON.stringify(
+                data.before
+              )}, after=${JSON.stringify(data.after)}`
+            );
```
nofir, line 65: i think we need to do a return here.

nofir (changes requested):

nofir (approved):

packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
tv3636: added the return

nofir (approved):

PR #5621, opened 2023-07-17
fix: handle invalid collectionsSetId on tokens API

tv3636:


packages/indexer/src/api/endpoints/tokens/get-tokens/v5.ts
```diff
@@ -846,6 +846,14 @@ export const getTokensV5Options: RouteOptions = {
       if (query.collectionsSetId) {
         const collectionsSetQueries = [];
         const collections = await CollectionSets.getCollectionsIds(query.collectionsSetId);
+
```
nofir, line 849: I see that in other places we have this logic:
`    if (query.collectionsSetId) {
      query.collection = await CollectionSets.getCollectionsIds(query.collectionsSetId);
      if (_.isEmpty(query.collection)) {
        throw Boom.badRequest(`No collections for collection set ${query.collectionsSetId}`);
      }
    }`

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -839,6 +839,13 @@ export const getTokensV6Options: RouteOptions = {
 
       // Break query into UNION of results for each collectionId for sets up to 20 collections
       if (query.collectionsSetId && collections.length <= 20) {
+        if (collections.length === 0) {
```
nofir, line 842: Same as above

packages/indexer/src/api/endpoints/tokens/get-tokens/v5.ts
nofir: maybe lets do the same for consistency

nofir (changes requested):

nofir (approved):

PR #5633, opened 2023-07-18
feat: Add sortDirection parameter

r3lays:


r3lays:
Waiting for confirmation that the `updated_at` index presumably supports the sortDirection for both `asc` and `desc`

packages/indexer/src/api/endpoints/transfers/get-transfers-bulk/v1.ts
```diff
@@ -201,12 +208,21 @@ export const getTransfersBulkV1Options: RouteOptions = {
             nft_transfer_events.batch_index DESC
         `;
       } else if (query.orderBy == "updated_at") {
-        baseQuery += `
+        if (query.sortDirection === "asc") {
+          baseQuery += `
+          ORDER BY
+            nft_transfer_events.updated_at ASC,
```
ipeleg, line 214: instead of the if you can inject the query.sortDirection into the string

```
baseQuery += `
          ORDER BY
            nft_transfer_events.updated_at ${query.sortDirection},
            nft_transfer_events.address ${query.sortDirection},
            nft_transfer_events.token_id ${query.sortDirection}
        `
```

packages/indexer/src/api/endpoints/transfers/get-transfers-bulk/v1.ts
ipeleg, line 1: Need also to update the continuation constructing in this line
https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/api/endpoints/transfers/get-transfers-bulk/v1.ts#L165

in a similar way to what is being done here
https://github.com/reservoirprotocol/indexer/blob/77d9f0a8daee50d0ed3cb95748b23ed270093b9c/packages/indexer/src/api/endpoints/tokens/get-tokens/v4.ts#L303

ipeleg (changes requested):

ipeleg:
@r3lays and if we're already at it, let's add the same logic to GET /transfers/v3

ipeleg (dismissed):

ipeleg (approved):

PR #5665, opened 2023-07-20
Tv/platf 1723 switch collections ws to use cdc event data

tv3636:


packages/indexer/src/jobs/websocket-events/collection-websocket-events-trigger-job.ts
```diff
@@ -1,27 +1,51 @@
...
+  day30_volume_change: number;
+  day1_floor_sell_value: string;
+  day7_floor_sell_value: string;
+  day30_floor_sell_value: string;
+  token_count: number;
+  owner_count: number;
+  floor_sell_id: string;
+  floor_sell_value: string;
+  floor_sell_maker: Buffer;
+  floor_sell_valid_between: string[];
```
nofir, line 36: does it actually come as array or a string?
`		"floor_sell_valid_between": "[\"2023-07-13 22:17:31+00\",\"2023-07-14 00:17:13+00\"]",`

tv3636: good catch. looking at the log it seems like everything ends up as either a string or a number

nofir (approved):

d3or (approved):

PR #5666, opened 2023-07-20
feat: use cdc event data for sale websocket

tv3636:


packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
```diff
@@ -44,36 +45,8 @@ export class SaleWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJobHandl
...
-          fill_events_2.marketplace_fee_bps,
-          fill_events_2.royalty_fee_breakdown,
-          fill_events_2.marketplace_fee_breakdown,
-          fill_events_2.paid_full_royalty,
-          fill_events_2.is_deleted,
-          fill_events_2.updated_at,
-          fill_events_2.created_at
+        SELECT          
+          currencies.decimals          
         FROM fill_events_2
```
nofir, line 50: i dont think you need to query the fill_events_2 table. just the currencies and tokens

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
tv3636: thanks, actually don't need currencies either, not sure why we were getting `decimals` since it's not used. updated the query and types

nofir (approved):

d3or (approved):

PR #5667, opened 2023-07-20
feat: use cdc event data for token ws

tv3636:


packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
```diff
@@ -60,46 +60,12 @@ if (config.doBackgroundWork && config.doWebsocketServerWork) {
...
-          t.rarity_score,
-          t.rarity_rank,
-          t.is_flagged,
-          t.last_flag_update,
-          t.last_flag_change,
           c.slug,
           (c.metadata ->> 'imageUrl')::TEXT AS collection_image,
-          t.supply,
-          t.remaining_supply
         FROM tokens t
```
nofir, line 69: Why is tokens table needed here? can't you use the collection_id that you get from event to just query the collections table?

nofir: if the collection_id is null just query the contracts table

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
```diff
@@ -118,77 +84,83 @@ if (config.doBackgroundWork && config.doWebsocketServerWork) {
...
-          contract: toBuffer(data.after.contract),
+          contract: data.after.contract,
           tokenId: data.after.token_id,
         });
 
         const r = rawResult[0];
 
-        const contract = fromBuffer(r.contract);
-        const tokenId = r.token_id;
+        const contract = fromBuffer(data.after.contract);
```
nofir, line 93: just to confirm, you get it as a buffer from the cdc event?

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
tv3636: it would be a string, will fix

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
```diff
@@ -60,135 +60,95 @@ if (config.doBackgroundWork && config.doWebsocketServerWork) {
...
-        t.floor_sell_currency_value,
-        t.normalized_floor_sell_id,
-        t.normalized_floor_sell_maker,
-        t.normalized_floor_sell_valid_from,
-        t.normalized_floor_sell_valid_to,
-        t.normalized_floor_sell_source_id_int,
-        t.normalized_floor_sell_value,
-        t.normalized_floor_sell_currency,
-        t.normalized_floor_sell_currency_value
+        const baseQuery = `
```
nofir, line 63: Potentially, there could be multiple collections with different images on the contract so i would do this:
1. check if the token has collecction_id (you get it from the cdc), if it does, query by contract and collection id (or just collection id)
2. otherwise, just query by contract (like you are doing now)

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-queue.ts
tv3636: added a condition to check collection id if it exists on `data.after`

nofir (approved):

d3or (approved):

PR #5678, opened 2023-07-22
feat: royalties improvement

lljxx1:


packages/indexer/src/migrations/1689907665189_fee-recipients.sql
```diff
@@ -0,0 +1,17 @@
+-- Up Migration
+
+CREATE TABLE "fee_recipients" (
```
georgeroman, line 3: Let's use the following schema instead:
```sql

CREATE TYPE "fee_kind_t" AS ENUM (
  'marketplace',
  'royalty'
);

CREATE TABLE "fee_recipients" (
  "address" BYTEA NOT NULL,
  "kind" "fee_kind_t" NOT NULL,
  "source_id" INT
);

ALTER TABLE "fee_recipients"
  ADD CONSTRAINT "fee_recipients_pk"
  PRIMARY KEY ("address", "kind");
```

georgeroman (changes requested):
Also, let's add an admin API for adding new fee recipients.

lljxx1:
yeah, it will automatically sync to database

PR #5692, opened 2023-07-24
fix: sort by tx and filter out ens + uniswap mints

lunarmayor:


packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -182,6 +182,12 @@ export enum TopSellingFillOptions {
   any = "any",
 }
 
+const trendingBlocklist = [
```
nofir, line 185: if its only relevant for eth mainnet, maybe have it in the network settings for the index:
https://github.com/reservoirprotocol/indexer/blob/8b624d6a6a4c33e6c850e4e3c43f13e795607d49/packages/indexer/src/config/network.ts#L199

nofir (changes requested):

packages/indexer/src/elasticsearch/indexes/activities/index.ts
lunarmayor: sorry took a bit but made the change

lunarmayor: @nofir good to merge?

nofir (approved):

packages/indexer/src/elasticsearch/indexes/activities/index.ts
nofir: approved

PR #5705, opened 2023-07-25
feat: foundation allowlist mint

lljxx1:


packages/indexer/src/orderbook/mints/calldata/detector/foundation.ts
```diff
@@ -1,21 +1,77 @@
...
+  getAllowlist,
+  allowlistExists,
+  createAllowlist,
+} from "@/orderbook/mints/allowlists";
 
 const STANDARD = "foundation";
 
+async function traceCreateSaleEvent(collection: string, maxRetry = 5, blockInterval = 10000) {
+  const face = new Interface([
+    `event CreateFixedPriceSale(
```
georgeroman, line 31: There's two events we need to listen to, `CreateFixedPriceSale` and `AddMerkleRootToFixedPriceSale`

packages/indexer/src/orderbook/mints/calldata/detector/foundation.ts
```diff
@@ -1,21 +1,77 @@
...
+  let merkleTreeUri: string | undefined;
+
+  for (let index = 0; index < maxRetry; index++) {
+    const createFixedPriceSaleLogs = await baseProvider.getLogs({
+      fromBlock: blockNumber - blockInterval * (index + 1),
+      toBlock: blockNumber - blockInterval * index,
+      topics,
+    });
+    if (createFixedPriceSaleLogs.length) {
+      const firstLog = createFixedPriceSaleLogs[0];
```
georgeroman, line 57: I think events are ordered ascending, right? Do we need to get the last one instead?

georgeroman (changes requested):
One additional comment is that we also want to listen to the two mentioned events in real-time so that once we detect one we'll update any associated mints as fast as possible

PR #5739, opened 2023-07-26
Midaswap

georgeroman:


packages/sdk/src/midaswap/order.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.chainId = chainId;
+
+    try {
+      this.params = normalize(params);
+    } catch {
+      throw new Error("Invalid params");
+    }
+  }
+
+  public static binToPriceFixed = (bin: number, decimal = 18, toFixedNumber = 18) => {
```
georgeroman, line 20: We should use BigNumber math for full precision.

packages/sdk/src/midaswap/order.ts
```diff
@@ -0,0 +1,62 @@
...
+    const powValue = bin - 8388608;
+    const b = Math.pow(10, 18 - decimal);
+    const price = (Math.pow(1.0001, powValue) * b).toFixed(toFixedNumber);
+    return String(price);
+  };
+
+  public static getSellPrice = (bin: number, freeRate = 0, royalty = 0) => {
```
georgeroman, line 27: We should use BigNumber math for full precision.

packages/sdk/src/midaswap/order.ts
```diff
@@ -0,0 +1,62 @@
...
+    const price = Order.binToPriceFixed(bin);
+    return parseEther(
+      (+price * (1 + (freeRate + royalty) / 10000)).toFixed(18).toString()
+    ).toString();
+  };
+
+  public static getBuyPrice = (bin: number, freeRate = 0, royalty = 0) => {
```
georgeroman, line 34: We should use BigNumber math for full precision.

packages/indexer/src/orderbook/orders/midaswap/index.ts
```diff
@@ -0,0 +1,1146 @@
...
+        bps: number;
+      }[] = [
+        {
+          kind: "marketplace",
+          recipient: pool.address,
+          bps: pool.freeRateBps,
+        },
+      ];
+
+      switch (orderParams.eventName) {
```
georgeroman, line 101: I'd suggest following all the other exchange implementations. Instead of having different logic based on the event that triggered an order update we should have generic logic which doesn't care about the triggering event but simply extracts the orders for any given pool / lp token id.

packages/sdk/src/midaswap/types.ts
```diff
@@ -0,0 +1,16 @@
+export type OrderParams = {
+  pair: string;
+  // Only relevant for listings
+  tokenX: string; // NFT
+  tokenY: string; // Token
+  lpTokenId: string;
+  pool: string; // `${pair}_${lpTokenId}`
+  tokenId?: string;
+  amount?: string;
+  extra: {
```
georgeroman, line 10: I refactored this so that we don't have any custom logic just for midaswap.

georgeroman:
Also, there are some offer tests that are failing.

packages/indexer/src/migrations/1687936565391_midaswap-pools.sql
```diff
@@ -0,0 +1,17 @@
+-- Up Migration
+
+CREATE TABLE "midaswap_pools" (
+  "address" BYTEA NOT NULL,
+  "nft" BYTEA NOT NULL, 
+  "token" BYTEA NOT NULL,
+  "free_rate_bps" NUMERIC(78, 0) NOT NULL,
```
georgeroman, line 7: How big are these bps going to be? Maybe worth changing the type to `INT` instead.

packages/sdk/src/midaswap/order.ts
LXVC: `bn("10001").pow(5).mul(1).div(bn(10000).pow(5))`, How to solve the problem that using div fractional parts will be ignored?

georgeroman: How are you calculating this in Solidity?

PR #5758, opened 2023-07-26
feat: increase limit to 100 on /owners/common-collections/v1

tv3636:


nofir (approved):
did you test the impact of the query with 100 limit?

tv3636:
> did you test the impact of the query with 100 limit?

yeah there's virtually no impact. however I think the request here may have been to increase the limit on # of owner addresses rather than results, waiting to hear back to confirm

PR #5779, opened 2023-07-27
feat: add contract kind and update metadata naming to match API

tv3636:


packages/indexer/src/jobs/websocket-events/collection-websocket-events-trigger-job.ts
```diff
@@ -103,6 +105,32 @@ export class CollectionWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJo
     const { data } = payload;
 
     try {
+      let contractKind = await redis.get(`collection-contract-kind:${data.after.contract}`);
```
nofir, line 108: can just be "contract-kind:${data.after.contract}", just if we ever use it for tokens

packages/indexer/src/jobs/websocket-events/collection-websocket-events-trigger-job.ts
```diff
@@ -149,12 +177,18 @@ export class CollectionWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJo
           slug: r.slug,
           name: r.name,
           metadata: {
-            ...metadata,
-            imageUrl: Assets.getLocalAssetsLink(metadata?.imageUrl),
+            image: Assets.getLocalAssetsLink(metadata?.imageUrl),
```
nofir, line 180: seem in the api we return the openseaVerificationStatus on the root level and not on metadata. maybe do the same and dont add all the fields from the metadata

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/collection-websocket-events-trigger-job.ts
tv3636: on the API we return them all at the root level, but if we do that here then we can't include metadata in `changed` right?

nofir: not sure i understand. 
also, imageUrl was already there, you cant change it to image, it might break for some users.

nofir (approved):

PR #5807, opened 2023-08-01
fix: add createdAt and updatedAt where missing

tv3636:


packages/indexer/src/jobs/websocket-events/top-bid-websocket-events-trigger-job.ts
```diff
@@ -48,6 +48,7 @@ export class TopBidWebSocketEventsTriggerJob extends AbstractRabbitMqJobHandler
                 orders.normalized_value,
                 orders.currency_normalized_value,               
                 orders.created_at,
+                orders.updated_at,
```
nofir, line 51: not needed here

nofir (approved):

PR #5818, opened 2023-08-02
feat: royalties improve

lljxx1:


packages/indexer/src/sync/events/handlers/royalties/calldata.ts
```diff
@@ -0,0 +1,43 @@
...
+    } else if (["fulfillAdvancedOrder", "fulfillOrder"].includes(funcName)) {
+      orders = [args[0]];
+    }
+
+    for (let index = 0; index < orders.length; index++) {
+      try {
+        const order = new Sdk.SeaportV15.Order(config.chainId, {
+          ...orders[index].parameters,
+          // the counter is missing
+          counter: 0,
```
georgeroman, line 31: We actually have the latest counter in the database, should use that instead

georgeroman (changes requested):

PR #5947, opened 2023-08-10
feat: add creator and token count to tokens API

tv3636:


packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1125,11 +1129,13 @@ export const getTokensV6Options: RouteOptions = {
               : undefined,
             media: r.media,
             kind: r.kind,
+            creator: r.creator ? fromBuffer(r.creator) : null,
```
nofir, line 1132: this should be on the collection obj

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1125,11 +1129,13 @@ export const getTokensV6Options: RouteOptions = {
...
             isFlagged: Boolean(Number(r.is_flagged)),
             lastFlagUpdate: r.last_flag_update ? new Date(r.last_flag_update).toISOString() : null,
             lastFlagChange: r.last_flag_change ? new Date(r.last_flag_change).toISOString() : null,
             supply: !_.isNull(r.supply) ? r.supply : null,
             remainingSupply: !_.isNull(r.remaining_supply) ? r.remaining_supply : null,
+            tokenCount: r.token_count,
```
nofir, line 1138: this should be on the collection obj

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -242,6 +242,8 @@ export const getTokensV6Options: RouteOptions = {
               name: Joi.string().allow("", null),
               image: Joi.string().allow("", null),
               slug: Joi.string().allow("", null),
+              creator: Joi.string().lowercase().pattern(regex.address).required(),
```
nofir, line 245: why is there required()? also, should allow null

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
tv3636: yeah not sure, fixed

nofir (approved):

PR #5948, opened 2023-08-10
feat: add filtering by creator on collections API

tv3636:


packages/indexer/src/api/endpoints/collections/get-collections/v6.ts
```diff
@@ -474,6 +475,10 @@ export const getCollectionsV6Options: RouteOptions = {
         query.contract = query.contract.map((contract: string) => toBuffer(contract));
         conditions.push(`collections.contract IN ($/contract:csv/)`);
       }
+      if (query.creator) {
```
nofir, line 478: what about adding an index?

tv3636: i commented in linear but basically this is pretty fast when filtering on just creator. i was thinking we can hold off adding an index until seeing if it's needed based on usage

nofir (approved):

PR #5978, opened 2023-08-14
feat: add migration for slug not null, remove slugify

d3or:


nofir (commented):
Why are there so many changes in the yarn.lock?

Other than that, looks good

PR #5991, opened 2023-08-14
Tv/platf 1720 switch bids and asks ws to use cdc event data

tv3636:


nofir (approved):

packages/indexer/src/utils/websockets/index.ts
```diff
@@ -0,0 +1,32 @@
+// Utility functions for parsing cdc event data
+
+export const formatValidBetween = (validBetween: string) => {
```
nofir, line 3: I would put this under "jobs/websocket-events/utils.ts". is not really relevant outside of it

nofir (changes requested):

nofir (approved):

PR #6021, opened 2023-08-17
Tv/platf 2080 backfill collection creator

tv3636:


packages/indexer/src/jobs/backfill/backfill-collections-creator-job.ts
```diff
@@ -0,0 +1,76 @@
...
+      ]);
+    }
+
+    if (result.length == limit) {
+      await this.addToQueue(result[result.length - 1].collection_id);
+    }
+  }
+
+  public async addToQueue(events: BackfillCollectionsCreatorJobPayload[]) {
+    if (!config.doBackgroundWork) {
```
nofir, line 64: doBackgroundWork is only to decide if to process jobs or not, not if they go in the queue

packages/indexer/src/jobs/backfill/backfill-collections-creator-job.ts
```diff
@@ -0,0 +1,76 @@
...
+    const { continuation } = payload;
+    const result = await idb.manyOrNone(
+      `
+        WITH x AS (
+          SELECT id AS collection_id
+          FROM collections c
+          WHERE creator IS NULL
+          ${continuation ? `AND id > '$/continuation/'` : ""}
+          ORDER BY id
+          LIMIT 100
```
nofir, line 34: you are not using the limit

packages/indexer/src/jobs/backfill/backfill-collections-creator-job.ts
```diff
@@ -0,0 +1,76 @@
...
+import { fromBuffer } from "@/common/utils";
+import { config } from "@/config/index";
+import { fetchCollectionMetadataJob } from "@/jobs/token-updates/fetch-collection-metadata-job";
+import { AbstractRabbitMqJobHandler, BackoffStrategy } from "@/jobs/abstract-rabbit-mq-job-handler";
+
+export type BackfillCollectionsCreatorJobPayload = {
+  continuation?: string;
+};
+
+export class BackfillCollectionsCreatorJob extends AbstractRabbitMqJobHandler {
```
nofir, line 11: Not a big deal but this might be better in bullmq as this is temporary job. cleaning the queues from rabbitmq is much more of a pain now. We also already have this job:
https://github.com/reservoirprotocol/indexer/blob/main/packages/indexer/src/jobs/backfill/backfill-refresh-collections-metadata.ts

maybe we can add a param the defines the logic like "missingCreator" and based on that change the logic of which collections to fetch. that way we have one job that can support different cases for the future. (then it might make sense to have it on rabbitmq as its something that will be used in the future when adding new types)

nofir (changes requested):

PR #6022, opened 2023-08-17
fix: check for tokenIdRange in collection when refreshing

tv3636:


packages/indexer/src/jobs/backfill/backfill-collections-split-collection.ts
```diff
@@ -0,0 +1,85 @@
...
+  concurrency = 1;
+  singleActiveConsumer = true;
+  useSharedChannel = true;
+  backoff = {
+    type: "exponential",
+    delay: 10000,
+  } as BackoffStrategy;
+  intervalInSeconds = 5;
+
+  protected async process(payload: BackfillCollectionsSplitCollectionJobPayload) {
```
nofir, line 24: I don't think the logic here will solve the issue .the issue is when trying to update the collection id for tokens:, [here](https://github.com/reservoirprotocol/indexer/blob/c83f747a35a4da76a4b7249ead632499e1cb1357/packages/indexer/src/api/endpoints/admin/post-refresh-collection.ts#L96).
You should move that logic here and do it by batches. I would call it something like "refresh-tokens-collection" or "refresh-tokens-collection-by-token-range", etc..

and we already do the token refresh by ["full-collection"](https://github.com/reservoirprotocol/indexer/blob/c83f747a35a4da76a4b7249ead632499e1cb1357/packages/indexer/src/api/endpoints/admin/post-refresh-collection.ts#L153), which already inside the job iterates the tokens by batches so there is not need to do it by "single token"

nofir (approved):

PR #6031, opened 2023-08-18
Ofir/platf 2050 return ask that sdk would use first

nofir:


packages/indexer/src/api/endpoints/orders/get-orders-asks/v5.ts
```diff
@@ -548,11 +571,19 @@ export const getOrdersAsksV5Options: RouteOptions = {
           if (query.normalizeRoyalties) {
             continuation = buildContinuation(
               rawResult[rawResult.length - 1].normalized_value ??
-                rawResult[rawResult.length - 1].price + "_" + rawResult[rawResult.length - 1].id
+                rawResult[rawResult.length - 1].price +
```
tv3636, line 574: shouldn't this be `rawResult[rawResult.length - 1].normalized_value`?

nofir:
```
rawResult[rawResult.length - 1].normalized_value ??
                rawResult[rawResult.length - 1].price
``` 

so if the normalized value is set, it will use it, otherwise it will fallback to price. No?

tv3636: ah you're right, formatting threw me off

tv3636 (approved):

PR #6069, opened 2023-08-23
fix: update backfill with startTimestamp

tv3636:


packages/indexer/src/jobs/backfill/backfill-sales-usd-price.ts
```diff
@@ -123,7 +124,7 @@ if (config.doBackgroundWork) {
         );
         for (const { tx_hash, log_index, batch_index, currency, price, usd_price } of results) {
           if (!usd_price) {
-            const prices = await getUSDAndNativePrices(fromBuffer(currency), price, timestamp, {
+            const prices = await getUSDAndNativePrices(fromBuffer(currency), price, endTimestamp, {
```
nofir, line 127: Not sure about this, would that give you the price for the fill date timestamp?

tv3636: i just mapped `timestamp` to `endTimestamp`, it gets updated with each subsequent job as `lastResult.timestamp` (line 165)

tv3636: i see what you mean though, i guess it should use the timestamp from the result to be sure

packages/indexer/src/jobs/backfill/backfill-sales-usd-price.ts
```diff
@@ -65,7 +66,7 @@ if (config.doBackgroundWork) {
         `,
         {
           limit,
-          timestamp,
+          endTimestamp,
```
nofir, line 69: you might want disable the logic below:
// Fix 1: Set the currency of old bids to WETH instead of ETH

packages/indexer/src/jobs/backfill/backfill-sales-usd-price.ts
```diff
@@ -121,7 +122,15 @@ if (config.doBackgroundWork) {
...
+          tx_hash,
+          log_index,
+          batch_index,
+          currency,
+          price,
+          usd_price,
+          timestamp,
+        } of results) {
           if (!usd_price) {
             const prices = await getUSDAndNativePrices(fromBuffer(currency), price, timestamp, {
```
nofir, line 135: Maybe add logic that only if the price is different from what it was before

nofir (changes requested):

nofir (dismissed):

nofir (approved):

PR #6071, opened 2023-08-23
feat: add flagged status to user tokens

tv3636:


packages/indexer/src/api/endpoints/tokens/get-user-tokens/v7.ts
```diff
@@ -158,6 +158,7 @@ export const getUserTokensV7Options: RouteOptions = {
               .allow(null)
               .description("No rarity rank for collections over 100k"),
             media: Joi.string().allow(null),
+            isFlagged: Joi.boolean().default(false),
```
nofir, line 161: lets make it similar to the tokens api:
            lastFlagUpdate: Joi.string().allow("", null),
            lastFlagChange: Joi.string().allow("", null),

nofir (changes requested):

nofir (approved):

PR #6100, opened 2023-08-25
feat: add collection data to top selling collection endpoint

pedromcunha:


packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v1.ts
```diff
@@ -96,51 +121,163 @@ export const getTopSellingCollectionsOptions: RouteOptions = {
...
+      period,
+      fillType,
+      limit,
+      includeRecentSales,
+      normalizeRoyalties,
+      useNonFlaggedFloorAsk,
+    } = request.query;
 
     try {
+      let startTime = Math.floor(new Date().getTime() / 1000) - 60 * 24 * 60;
```
lunarmayor, line 136: can we save this to a variable `Math.floor(new Date().getTime() / 1000)` seems to be repeated a lot here

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -205,7 +206,7 @@ const mapBucketToCollection = (bucket: any, includeRecentSales: boolean) => {
     id: collectionData?.id,
     name: collectionData?.name,
     image: collectionData?.image,
-    primaryContract: collectionData?.contract,
+    primaryContract: data?.contract,
```
lunarmayor, line 209: oh interesting guess this was never working?

pedromcunha: yeah it wasn't working, I patched it. We were just grabbing it from the wrong node.

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v1.ts
pedromcunha: done, good point!

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v1.ts
```diff
@@ -14,9 +14,9 @@ import {
 
 import { getJoiPriceObject, JoiPrice } from "@/common/joi";
 
-const version = "v5";
```
lunarmayor, line 17: oops 😅

lunarmayor (approved):
looks good to me

PR #6122, opened 2023-08-29
fix: update chain stats query on base and optimism

tv3636:


lunarmayor (approved):
looks good to me! thanks for fixing

PR #5121, opened 2023-06-14
feat/ditto-protocol

KjetilVaa:
Incremental PR to integrate Dittoswap into Reservoir.

There are three major steps to complete you integration:
1. SDK integration
   - definition for the native order format of the protocol  <--- currently here
    - smart contract module for filling (together with tests)
    - router code for handling buying and selling (see sdk/router/v6/router.ts)
3. Indexer integration
4. Router integration

PR #6124, opened 2023-08-29
feat: ditto router module and sdk

relyt29:
Hi folks,

We have decided to parse the changes required to add dittohq.xyz integration to reservoir in two smaller more manageable and reviewable PRs: one which does the on-chain settlement layer (the router module to trade with Ditto + sdk support for settlement) and one which adds the indexer support.

Right now the indexer efforts are focused on #5121 (CC: @KjetilVaa). This PR is for the settlement layer component as well as javascript support for trading on Ditto via the reservoir sdk.

This PR adds DittoModule.sol for the reservoir router module to trade with Ditto, tests for that functionality under `packages/contracts/test/router/ditto` and then adds an sdk module under `packages/sdk/src/ditto` with tests for the sdk module under `packages/contracts/test/sdk/ditto`

1. I wasn't sure if  `packages/contracts/test/sdk/ditto` was the right place to test sdk support, and if my contributions to the sdk are functional so that e.g. users of ReservoirKit UI can settle trades on Ditto now using the code in the `packages/sdk/src/ditto` folder. Are there further unit tests that need to be written to test the full end-to-end flow against Ditto?
2. We based our reservoir integration code off of sudoswap v1, which appears to not use a reservoir module but instead [route transactions through sudoswap's royalty router directly](https://github.com/reservoirprotocol/indexer/blob/8833a9f21c7fdd97fb96ef2cedab44afca286c3b/packages/sdk/src/sudoswap/router.ts#L20) bypassing the reservoir smart contracts. Is that correct? Shouldn't the reservoir sdk use the reservoir router smart contracts for execution? Should this PR changed so that the reservoir sdk uses the reservoir router ditto module instead?

Other than those two questions, please advise if there are any additional tasks required to get this PR merged into reservoir's codebase.

Thanks!

### DIFF
--- a/packages/contracts/contracts/router/modules/exchanges/DittoModule.sol
+++ b/packages/contracts/contracts/router/modules/exchanges/DittoModule.sol
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {BaseExchangeModule} from "./BaseExchangeModule.sol";
+import {BaseModule} from "../BaseModule.sol";
+import {IDittoPool} from "../../../interfaces/IDittoPool.sol";
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {ERC20} from "solmate/src/tokens/ERC20.sol";
+import {SafeTransferLib} from "solmate/src/utils/SafeTransferLib.sol";
+
+struct DittoOrderParams {
+  uint256[] nftIds;
+  bytes swapData;
+}
+
+contract DittoModule is BaseExchangeModule {
+  using SafeTransferLib for ERC20;
+
+  // --- Constructor ---
+  constructor(address owner, address router) BaseModule(owner) BaseExchangeModule(router) {}
+
+  function poolTransferNftFrom(
+    IERC721 nft, 
+    address from, 
+    address to, 
+    uint256 id
+  ) 
+  external 
+  {
+      // transfer NFTs to pool
+      nft.transferFrom(from, to, id);
+  }
+
+  function poolTransferErc20From(
+    ERC20 token,
+    address from,  
+    address to,
+    uint256 amount
+  ) 
+  external 
+  virtual 
+  {
+    // transfer tokens to txn sender
+    token.safeTransferFrom(from, to, amount);
+  }
+
+  // --- Multiple ERC20 listing ---
+
+  function buyWithERC20(
+    IDittoPool[] calldata pairs,
+    DittoOrderParams[] calldata orderParams,
+    ERC20ListingParams calldata params,
+    Fee[] calldata fees
+  )
+  external
+  nonReentrant
+  refundERC20Leftover(params.refundTo, params.token)
+  chargeERC20Fees(fees, params.token, params.amount)
+  {
+    uint256 pairsLength = pairs.length;
+    for (uint256 i; i < pairsLength; ) {
+
+      // Execute fill
+      IDittoPool.SwapTokensForNftsArgs memory args = IDittoPool.SwapTokensForNftsArgs({
+        nftIds: orderParams[i].nftIds,
+        maxExpectedTokenInput: params.amount,
+        tokenSender: params.fillTo,
+        nftRecipient: params.fillTo,
+        swapData: orderParams[i].swapData
+      }); 
+
+      pairs[i].swapTokensForNfts(args);
+
+      unchecked {
+        ++i;
+      }
+    }
+  }
+
+  // --- Single ERC721 offer ---
+
+  function sell(
+    IDittoPool pool,
+    DittoOrderParams calldata orderParams,
+    uint256[] calldata lpIds,
+    bytes calldata permitterData,
+    uint256 minOutput,
+    OfferParams calldata params,
+    Fee[] calldata fees
+  ) external nonReentrant {
+  
+      IERC20 token = pool.token();
+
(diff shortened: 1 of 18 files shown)


georgeroman:
Responding to your questions:
1. At the moment the SDK is only used for testing purposes, so no need to have tests for that. However, we need tests for the router module (eg. https://github.com/reservoirprotocol/indexer/tree/main/packages/contracts/test/router/sudoswap-v2)
2. As mentioned in 1, the SDK is only used for testing purposes. The actual calldata generation which is used in the execute APIs is below. That part uses the router module for filling.
https://github.com/reservoirprotocol/indexer/blob/8833a9f21c7fdd97fb96ef2cedab44afca286c3b/packages/sdk/src/router/v6/router.ts#L1928-L1992

relyt29:
Thanks those answers are helpful.
> 1. At the moment the SDK is only used for testing purposes, so no need to have tests for that. However, we need tests for the router module

Ok so we have some tests in `packages/contracts/test/router/ditto/listings.test.ts` that should cover that. Let me know if those are not adequate. I removed the tests for the "sdk" that I had prior living in `packages/contracts/test/sdk`

> 2. As mentioned in 1, the SDK is only used for testing purposes. The actual calldata generation which is used in the execute APIs is below. That part uses the router module for filling.

Ok so I added changes to `indexer/packages/sdk/src/router/v6/router.ts` - does this mean that with those changes to that specific file, clients using the Reservoir UI Kit package will be able to settle properly formatted Ditto Orders on chain using the reservoir ditto module? If so, how can I test that works?

Please let me know if there's anything else I can do to push this PR forward / closer to merging into main

packages/contracts/contracts/router/modules/exchanges/DittoModule.sol
```diff
@@ -0,0 +1,121 @@
...
+pragma solidity ^0.8.9;
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {BaseExchangeModule} from "./BaseExchangeModule.sol";
+import {BaseModule} from "../BaseModule.sol";
+import {IDittoPool} from "../../../interfaces/IDittoPool.sol";
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {ERC20} from "solmate/src/tokens/ERC20.sol";
```
georgeroman, line 11: If possible, let's use `openzeppelin` and avoid adding a new dependency.

packages/sdk/src/ditto/addresses.ts
```diff
@@ -0,0 +1,36 @@
+import { ChainIdToAddress, Network } from "../utils";
+
+export const DittoPoolFactory: ChainIdToAddress = {
+  [Network.Ethereum]: "TBD",
+  [Network.EthereumGoerli]: "0xF316cfF1170e9f04e8EdB4f67D9D5BCdD84F142b",
```
georgeroman, line 5: All addresses should be lowercase.

packages/sdk/src/ditto/order.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }
+}
+
+const normalize = (order: Types.OrderParams): Types.OrderParams => {
+  // Perform some normalization operations on the order:
+  // - convert bignumbers to strings where needed
+  // - convert strings to numbers where needed
+  // - lowercase all strings
+  return {
```
georgeroman, line 24: I see some `BigNumber` values here, we should convert those all to `string`.

georgeroman (changes requested):

packages/sdk/src/ditto/order.ts
relyt29: is that because they go into the orders postgres database, when the indexer component is ready, and the database stores them as strings? Because from a typing enforcement / typescript / javascript / on-chain contract call perspective BigNumber is the more correct choice

That said it's your codebase i'll change them to strings, just trying to understand why

georgeroman: Exactly, we will store this data in a db so we try to avoid object types.

relyt29:
addressed review comments, please let me know if anything else is needed:

For testing I am using an .env under `packages/contracts` like so:

```
RPC_URL="https://eth-goerli.g.alchemy.com/v2/APIKEY"
CHAIN_ID=5
ALCHEMY_KEY="APIKEY"
DEPLOYER_PK="APRIVATEKEYFORTESTING"
BLOCK_NUMBER="9574400"
ETHERSCAN_API_KEY="APIKEYFORETHERSCAN"
REPORT_GAS=0
```

and run the tests using e.g.

```
yarn test test/router/ditto/listings.test.ts
yarn test test/router/ditto/offers.test.ts
yarn test test/router/ditto/listings-oracle.test.ts
```

let me know if there's anything more I can do to push this ball further forward

packages/contracts/contracts/router/modules/exchanges/DittoModule.sol
```diff
@@ -0,0 +1,121 @@
...
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {BaseExchangeModule} from "./BaseExchangeModule.sol";
+import {BaseModule} from "../BaseModule.sol";
+import {IDittoPool} from "../../../interfaces/IDittoPool.sol";
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {SafeTransferLib} from "solmate/src/utils/SafeTransferLib.sol";
```
georgeroman, line 12: The duplicate module dependency applies to this as well. Can we replace this with something else?

PR #6166, opened 2023-09-01
feat: add collectionNewContractDeployedJob

d3or:


packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -0,0 +1,99 @@
...
+      logger.error(this.queueName, `Missing contract or deployer`);
+      return;
+    }
+
+    // get the type of the collection, either ERC721 or ERC1155. if it's not one of those, we don't care
+    // get this from the contract itself
+    const collectionKind = await detectTokenStandard(contract);
+
+    switch (collectionKind) {
+      case "ERC721":
```
nofir, line 32: What about ERC721C that was added for the payment processor protocol?

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -0,0 +1,99 @@
...
+export class CollectionNewContractDeployedJob extends AbstractRabbitMqJobHandler {
+  queueName = "collection-new-contract-deployed";
+  maxRetries = 10;
+  concurrency = 10;
+  persistent = false;
+
+  protected async process(payload: CollectionContractDeployed) {
+    const { contract, deployer } = payload;
+
+    if (!contract || !deployer) {
```
nofir, line 22: Why is this an OR? isn't the contract required for the logic?

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -0,0 +1,99 @@
...
+        {
+          address: contract,
+          kind: collectionKind.toLowerCase(),
+          symbol: contractName.symbol || null,
+          name: contractName.name || null,
+        }
+      ),
+      idb.none(
+        `
+              INSERT INTO collections (
```
nofir, line 72: What would happen when the first token is minted? today we have to logic to create the collection once we get the collection metadata, would that not happen anymore?

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
d3or: Every ERC-721 (nft) compliant contract must implment ERC721 and ERC165 interfaces, which can be checked via the identifier `0x80ac58cd`

More info here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md#specification

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
d3or: If either are missing that means that something is wrong with the trace / we are missing some data. Both contract and deployer should never be null

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
d3or: We should still keep that for now. Once this goes live there will be collections that have already been deployed but have no mints yet, which we wont catch with this job and its only being called from real time.

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -0,0 +1,106 @@
...
+                $/name/,
+                $/contract/,
+                $/creator/
+              ) ON CONFLICT DO NOTHING
+            `,
+        {
+          id: contract,
+          name: contractName || null,
+          contract: toBuffer(contract),
+          creator: deployer,
```
georgeroman, line 95: Same here, I assume `creator` is a `BYTEA`

packages/indexer/src/jobs/token-updates/mint-queue-job.ts
```diff
@@ -56,7 +57,26 @@ export class MintQueueJob extends AbstractRabbitMqJobHandler {
         }
       );
 
+      let isFirstToken = false;
```
nofir, line 60: probably can be inside the if

nofir (dismissed):

nofir (approved):

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -0,0 +1,115 @@
...
+        ON CONFLICT (address) DO UPDATE SET
+          symbol = EXCLUDED.symbol,
+          name = EXCLUDED.name
+      `,
+        {
+          address: toBuffer(contract),
+          kind: collectionKind.toLowerCase(),
+          symbol: symbol || null,
+          name: name || null,
+          created_at: payload.blockTimestamp ? new Date(payload.blockTimestamp * 1000) : null,
```
nofir, line 78: i think created_at should always be the db insert timestamp and we should have another field to represent the contract deployment timestamp. i think it will be helpful especially if we do a backfill

nofir (changes requested):

nofir (approved):

PR #6176, opened 2023-09-01
feat: add sortBy updatedAt for collections API

tv3636:


packages/indexer/src/api/endpoints/collections/get-collections/v6.ts
```diff
@@ -661,6 +673,7 @@ export const getCollectionsV6Options: RouteOptions = {
             id: r.id,
             slug: r.slug,
             createdAt: new Date(r.created_at).toISOString(),
+            updatedAt: new Date(r.updated_at).toISOString(),
```
nofir, line 676: I remember we had an issue with the sort when not converting this to epoch timestamp. seems like its already done wrong for createdAt but i think worth reviewing and maybe fixing both (might need a new version)

nofir (approved):

PR #6186, opened 2023-09-05
feat: add originated_at for caviar orders

tv3636:


packages/indexer/src/orderbook/orders/caviar-v1/index.ts
```diff
@@ -234,6 +234,7 @@ export const save = async (orderInfos: OrderInfo[]): Promise<SaveResult[]> => {
               currency_normalized_value: normalizedValue.toString(),
               block_number: orderParams.txBlock ?? null,
               log_index: orderParams.logIndex ?? null,
+              originated_at: validFrom,
```
nofir, line 237: Shouldn't this be based on the txTimestamp?

tv3636: it is, see [line 202](https://github.com/reservoirprotocol/indexer/pull/6186/files#diff-a04a3aa09a8686247287afc177a110568d4f5d36ea903a6ec5a7a2ea081e81e8R202)

nofir: yes, my bad

nofir (approved):

PR #6204, opened 2023-09-06
fix: move sales latency log to cdc topic

tv3636:


nofir (approved):

packages/indexer/src/jobs/cdc/topics/indexer-fill-events.ts
```diff
@@ -22,6 +23,19 @@ export class IndexerFillEventsHandler extends KafkaEventHandler {
       },
       eventKind: WebsocketEventKind.SaleEvent,
     });
+
+    logger.info(
+      "sales-latency",
+      JSON.stringify({
+        latency: new Date(payload.after.created_at).getTime() / 1000 - payload.after.timestamp,
```
nofir, line 30: should we do a math.min here? not sure if it matters

PR #6209, opened 2023-09-07
feat: gas estimation

lljxx1:


georgeroman:
@lljxx1 I see we're still relying on `eth_estimateGas` which as I've mentioned we probably want to avoid for now (due to multi-transaction complexity). Let's move to a simpler design:
- In the router each transaction should return a list of tags, as you designed it so far. However, I feel like those tags should be key/value pairs instead of simple keys. Initially probably the following tags are enough:
  - `kind`: `approval` / `sale` / `mint` / `swap` (this is mandatory, while everything below is optional)
  - `approvals`: `number of approvals triggered`
  - `listings:PROTOCOL`: `number of listings filled for PROTOCOL`
  - `bids:PROTOCOL`: `number of bids filled for PROTOCOL`
  - `mints`: `number of mints filled`
  - `swaps`: `number of swaps triggered`
  - `fees-on-top`: `number of fees on top`
- Each transaction kind gets assigned a base fee (eg. `approval = 50000`, `sale = 100000`, `mint = 40000`, `swap = 100000`) and then each tag adds to this base fee (eg. `listings:SEAPORT: 3` would add `3 * gas cost to fill on Seaport`). This per tag gas cost can simply be hardcoded for now in a table like below. We can also have default values for protocols for which we don't know the cost (eg. a `listings:*` entry would cover anything that didn't match any previous rules).
```json
{
  "listings:seaport-v1.5": 30000,
  "listings:*": 40000,
  "swaps": 50000,
}
```

PR #6266, opened 2023-09-12
fix: deprecate v1 top selling collections

jamiechicago312:
Added tag: "x-deprecated"
Set: "hapi-swagger": {deprecated: true,}


lunarmayor (approved):
looks good. thanks!

PR #6273, opened 2023-09-12
Tv/platf 2051 unknown sources default to reservoirtools

tv3636:


nofir (changes requested):
in addition to the other comments, i think you missed the logic in the data export to dune, for example: 
https://github.com/reservoirprotocol/indexer/blob/fcf0f25cb8dd30214c35caf8326eaefefacec4f8/packages/indexer/src/jobs/data-export/data-sources/sales.ts#L68

packages/indexer/src/models/sources/index.ts
```diff
@@ -551,4 +530,32 @@ export class Sources {
       }
     }
   }
+
+  public getFullSourceObject(source: SourcesEntity | undefined): {
```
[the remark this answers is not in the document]
tv3636: in that case we'll have duplicate logic across all the older APIs that use source but not through the common joi objects, is that ok?

packages/indexer/src/models/sources/index.ts
```diff
@@ -272,11 +254,9 @@ export class Sources {
     tokenId?: string,
     optimizeCheckoutURL = false
   ): SourcesEntity | undefined {
-    let sourceEntity: SourcesEntity;
+    let sourceEntity: SourcesEntity | undefined;
```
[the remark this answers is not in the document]
tv3636: just curious, if we're making this change across all APIs including older versions, when would we ever still want to use the default?

packages/indexer/src/models/sources/index.ts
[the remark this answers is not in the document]
nofir: we can't add the common joi object in old apis?

packages/indexer/src/models/sources/index.ts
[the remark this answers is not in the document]
nofir: i don't think there is a usecase now but there might be one in the future (or we change our minds..)

packages/indexer/src/models/sources/index.ts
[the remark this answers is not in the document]
nofir: you're already doing ```source: sources.getSourceObject(source),``` Im just saying that the method here should be on the joi class as they are dedicated per external formatting

packages/indexer/src/jobs/data-export/data-sources/ask-events.ts
```diff
@@ -57,7 +57,9 @@ export class AskEventsDataSource extends BaseDataSource {
         quantity_remaining: Number(r.order_quantity_remaining),
         valid_from: r.valid_from ? Number(r.valid_from) : null,
         valid_until: r.valid_until ? Number(r.valid_until) : null,
-        source: sources.get(r.order_source_id_int)?.name,
+        source: sources.get(r.order_source_id_int)
```
nofir, line 60: instead of doing the get twice, just initialize a source param and do the iff with it (for all below files too)

nofir (changes requested):

packages/indexer/src/api/endpoints/orders/get-orders-bids/v1.ts
```diff
@@ -362,10 +362,10 @@ export const getOrdersBidsV1Options: RouteOptions = {
           validUntil: Number(r.valid_until),
           metadata: r.metadata,
           source: {
-            id: source?.address,
-            name: source?.getTitle(),
-            icon: source?.getIcon(),
-            url: source?.metadata.url,
+            id: source?.address ?? null,
```
nofir, line 365: just to confirm, will it now return null instead of undefined (property wont be there)?

packages/indexer/src/api/endpoints/sources/get-sources/v1.ts
```diff
@@ -91,14 +91,14 @@ export const getSourcesV1Options: RouteOptions = {
...
           return {
-            id: source.address ?? undefined,
-            name: source?.getTitle(),
-            description: source?.metadata.description,
-            socialImage: source?.metadata.socialImage,
-            twitterUsername: source?.metadata.twitterUsername,
-            icon: source?.getIcon(),
-            domain: source.domain ?? undefined,
-            tokenUrl: sources.getTokenUrl(source),
+            id: source.address ?? null,
```
nofir, line 94: why are the changes here needed?

packages/indexer/src/api/endpoints/orders/get-orders-bids/v1.ts
tv3636: yes exactly

packages/indexer/src/api/endpoints/sources/get-sources/v1.ts
tv3636: for consistency i was updating anywhere we return source to default to null rather than undefined, but if this case is separate i can undo the change

nofir: I dont think you need to make any changes here

packages/indexer/src/api/endpoints/orders/get-orders-bids/v1.ts
nofir: remind me, why are we doing it?

nofir (dismissed):

nofir (dismissed):

nofir (approved):

PR #6282, opened 2023-09-13
deor/platf 2195 move token collection metdata fetch simplehash opensea

d3or:
- feat: create abstract base provider + add opensea base provider
- feat: updates
- feat: add simplehash metadata provider
- feat: add rarible metadata provider
- feat: add soundxyz provider
- feat: update abstract base provider impl
- feat: add CenterdevMetadataProvider
- feat: add onchain metadata provider
- feat: metadata provider


packages/indexer/src/config/index.ts
```diff
@@ -84,12 +84,19 @@ export const config = {
...
+  openseaSlugBaseUrl: String(process.env.OPENSEA_SLUG_BASE_URL),
+  openseaBaseUrlAlt: String(process.env.OPENSEA_BASE_URL_ALT),
   openSeaApiUrl: String(process.env.OPENSEA_API_URL || ""),
 
   openSeaCrossPostingApiKey: String(
     process.env.OPENSEA_CROSS_POSTING_API_KEY || process.env.OPENSEA_API_KEY
   ),
 
+  simplehashApiKey: String(process.env.SIMPLEHASH_API_KEY),
+  centerdevApiKey: String(process.env.CENTERDEV_API_KEY),
```
nofir, line 98: I did not mention it but i think we can deprecate centerdev, we are not using it

packages/indexer/src/metadata/providers/abstract-base-metadata-provider.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  async getTokensMetadataBySlug(
+    slug: string,
+    continuation: string
+  ): Promise<TokenMetadataBySlugResult> {
+    // handle universal extend/custom logic here
+    return this._getTokensMetadataBySlug(slug, continuation);
+  }
+
+  normalizeMetadata = (collection: Collection): Metadata => {
```
nofir, line 31: small but this can be maybe in a utils class.

packages/indexer/src/metadata/providers/abstract-base-metadata-provider.ts
```diff
@@ -0,0 +1,184 @@
...
+  protected abstract _getTokensMetadata(
+    tokens: { contract: string; tokenId: string }[]
+  ): Promise<TokenMetadata[]>;
+
+  protected abstract _getTokensMetadataBySlug(
+    contract: string,
+    slug: string,
+    continuation?: string
+  ): Promise<TokenMetadataBySlugResult>;
+}
```
nofir, line 184: notice that you have a misspell in some of the files "provicer" instead of "provider"

packages/indexer/src/metadata/providers/simplehash-metadata-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+        name: data.collection.name,
+        community: null,
+        metadata: normalizeMetadata(data.collection),
+        contract,
+        tokenIdRange: null,
+        tokenSetId: `contract:${contract}`,
+        creator: _.toLower(data.contract.deployed_by),
+      };
+    } catch (error) {
+      logger.error(
```
nofir, line 49: does it make sense to put this in the abstract / base class for collection logic?

nofir: and only disable it for onchain provider

packages/indexer/src/metadata/providers/simplehash-metadata-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+      });
+
+    return data.nfts.map(this.parse).filter(Boolean);
+  }
+
+  async _getTokensMetadataBySlug(): Promise<TokenMetadataBySlugResult> {
+    throw new Error("Method not implemented.");
+  }
+
+  parse = (asset: any) => {
```
nofir, line 108: I think this should be an abstract method in the base class and then implemented in each provider class. at the end the response format should always be identical
OR 
it can be another class (BaseParser), similar to how it is today in the metdata but i actually think it makes more sense to keep it in the provider class

packages/indexer/src/metadata/providers/simplehash-metadata-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+      let slug = null;
+      if (_.isArray(data.collection.marketplace_pages)) {
+        for (const market of data.collection.marketplace_pages) {
+          if (market.marketplace_id === "opensea") {
+            slug = slugify(market.marketplace_collection_id, { lower: true });
+          }
+        }
+      }
+
+      return {
```
nofir, line 37: does it make sense to also take this logic into a parseCollection method?

packages/indexer/src/config/index.ts
```diff
@@ -84,12 +84,23 @@ export const config = {
 
   looksRareApiKey: String(process.env.LOOKSRARE_API_KEY),
   openSeaApiKey: String(process.env.OPENSEA_API_KEY),
+  openSeaSlugApiKey: String(process.env.OPENSEA_SLUG_API_KEY),
```
nofir, line 87: I think we can simplify this, it was using our old opensea proxy, lets discuss

packages/indexer/src/metadata/providers/opensea-metadata-provider.ts
```diff
@@ -88,16 +88,17 @@ class OpenseaMetadataProvider extends AbstractBaseMetadataProvider {
...
-              url,
-              "X-API-KEY": config.openSeaApiKey.trim(),
-              Accept: "application/json",
-            }
-          : {
-              Accept: "application/json",
-            },
+      .get(!this.isOSTestnet() ? config.openSeaBaseUrlAlt || url : url, {
+        headers:
+          config.chainId === 1
```
nofir, line 93: should be !this.isOSTestnet()

packages/indexer/src/metadata/providers/opensea-metadata-provider.ts
```diff
@@ -131,13 +132,14 @@ class OpenseaMetadataProvider extends AbstractBaseMetadataProvider {
 
     const url = `${
       config.chainId === 1
-        ? config.openseaSlugBaseUrl || "https://api.opensea.io"
```
nofir: i think this can be like the tokens:
    const url = `${
      !this.isOSTestnet() ? "https://api.opensea.io" : "https://testnets-api.opensea.io"
    }/api/v1/assets?${searchParams.toString()}`;

and then you dont need openseaSlugBaseUrl

nofir (changes requested):

packages/indexer/src/config/index.ts
```diff
@@ -86,6 +86,9 @@ export const config = {
   openSeaApiKey: String(process.env.OPENSEA_API_KEY),
   openSeaApiUrl: String(process.env.OPENSEA_API_URL || ""),
 
+  openSeaSlugApiKey: String(process.env.OPENSEA_SLUG_API_KEY),
```
nofir, line 89: 1. i think we need another dedicated one for the token flow (instead of using openSeaApiKey).
2. maybe we can make the naming a bit clear: (not critical)
- openseaTokenMetadataApiKey, openseaTokenMetadataBySlugApiKey, openseaCollectionMetadataApiKey

nofir (changes requested):

packages/indexer/src/jobs/metadata-index/metadata-process-job.ts
```diff
@@ -55,7 +55,7 @@ export class MetadataIndexProcessJob extends AbstractRabbitMqJobHandler {
 
     const results = await Promise.all(
       refreshTokensChunks.map((refreshTokensChunk) =>
-        MetadataApi.getTokensMetadata(
+        MetadataProviderRouter.getTokensMetadata(
```
nofir, line 58: Not sure if really matters but not sure if its cleaner to not return the axios response as this might not be an api call anymore (on chain) see below:
if (error.response?.status === 429) {

nofir: but this can be cleanup when we do the on chain story so we can this out the door

nofir (approved):

PR #6301, opened 2023-09-14
Tv/platf 1888 investigate submitting custom metrics to dd instead of using

tv3636:


tv3636:
closing since we aren't going to use custom metrics for now

PR #6322, opened 2023-09-18
Index foundation

zSlammin:
Indexes foundation shared contract by creator address, by updating slug to include creatorAddress and by updating creator field for collecton.


jamiechicago312:
We merged it in here, https://github.com/reservoirprotocol/indexer/pull/6419

We wanted to add a test case to confirm it works

PR #6329, opened 2023-09-18
feat: add redis cache to top selling collections

lunarmayor:


lunarmayor:
Definitely need to think about this more holistically but in the short term, I think this unblocks Magic Eden. It creates a job every 5 minutes to recalculate the top selling collections for the longer time periods and then serves them from a redis cache instead of calculating on the fly

nofir (approved):

PR #6373, opened 2023-09-20
feat: add partialBidSupported and traitBidSupported

pedromcunha:
Note: need to verify with george if traitBidding is supported on these exchanges (xy2y, looks-rare-v2, blur, seaport)


georgeroman:
@pedromcunha Trait bidding is only supported on Seaport, partial bidding is only supported on Seaport and Blur

georgeroman:
Looks good!

georgeroman (approved):

PR #6390, opened 2023-09-21
feat: add chain property

r3lays:


nofir:
@r3lays what is the goal of this change? there is no cross chain indexer right now so this seems a bit redundant

nofir (approved):

PR #6395, opened 2023-09-21
Lunarmayor/cache top selling collections

lunarmayor:


packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
```diff
@@ -33,11 +36,14 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
   },
   validate: {
     query: Joi.object({
-      period: Joi.string().valid("5m", "30m", "1h", "6h", "24h").default("24h"),
+      period: Joi.string()
+        .valid("5m", "10m", "30m", "1h", "6h", "1d", "24h")
```
pedromcunha, line 40: isn't 1d and 24h the same?

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
```diff
@@ -247,8 +262,9 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
         });
       }
       const sources = await Sources.getInstance();
+
       const collections = await Promise.all(
-        responses.map(async (response) => {
+        responses.map(async (response: any) => {
```
pedromcunha, line 267: Why any? Shouldn't we strive for type accuracy.

pedromcunha (changes requested):

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
lunarmayor: It is but I told ME that i would
allow it as an alias. Probably a better way to do this but didn't have time to figure it out

PR #6397, opened 2023-09-22
feat: permit bidding

lljxx1:


packages/indexer/src/migrations/1693809221377_permit-biddings.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE "orders" ADD COLUMN "permit_id" TEXT;
+
+ALTER TABLE "permit_biddings"
+  ADD CONSTRAINT "permit_biddings_pk"
+  PRIMARY KEY ("id");
+
+CREATE INDEX "permit_biddings_token_spender_owner_nonce_deadline_full_index"
+  ON "permit_biddings" ("token", "spender", "owner", "nonce", "deadline");
+
+CREATE INDEX "orders_permit_id_index"
```
georgeroman, line 24: The `orders` table is way too big to add a new column and index. Instead we should be storing the `permitId` in the `rawData` and use an existing index to query orders that were affected by permit changes (see the `order-updates-by-maker` job which queries by `maker` which is what we should be doing as well).

packages/indexer/src/migrations/1693809221377_permit-biddings.sql
```diff
@@ -0,0 +1,32 @@
+-- Up Migration
+
+CREATE TABLE "permit_biddings" (
```
georgeroman, line 3: Let's rename this to `permits` and make the following changes:
- add `kind` field (ideally this table will handle Permit2 as well): initially the only allowed value would be `eip2612` (https://eips.ethereum.org/EIPS/eip-2612), but we will probably support `permit2` soon
- add an `index` field (which is relevant for the batch permit of Permit2): we can represent a batch permit by having multiple rows with the same `id` but different `index`

packages/indexer/src/migrations/1693809221377_permit-biddings.sql
```diff
@@ -0,0 +1,32 @@
...
+    "id" TEXT NOT NULL,
+    "token" BYTEA NOT NULL,
+    "owner" BYTEA NOT NULL, 
+    "spender" BYTEA NOT NULL, 
+    "value" NUMERIC(78, 0),
+    "nonce" NUMERIC(78, 0),
+    "deadline" NUMERIC(78, 0),
```
georgeroman, line 10: This can be an `INT`

packages/indexer/src/migrations/1693809221377_permit-biddings.sql
```diff
@@ -0,0 +1,32 @@
...
+    "signature" TEXT NOT NULL,
```
georgeroman, line 11: This should be `BYTEA`

georgeroman (changes requested):

packages/contracts/contracts/misc/PermitProxy.sol
```diff
@@ -92,4 +103,25 @@ contract PermitProxy is ERC2771Context, ReentrancyGuard {
 
     ROUTER.execute(executionInfos);
   }
+
+  function bulkPermit(
```
georgeroman, line 107: Instead of adding a new method, which requires a redeployment across all chains, we can reuse the `transferWithExecute` method with empty `transfers`.

packages/contracts/contracts/misc/PermitProxy.sol
```diff
@@ -35,6 +35,17 @@ contract PermitProxy is ERC2771Context, ReentrancyGuard {
     Transfer[] transfers;
   }
 
+  struct PermitApproval {
```
georgeroman, line 38: We don't need this if reusing the `transferWithExecute` method.

PR #6418, opened 2023-09-25
feat: add one minute cache to top selling endpoint

lunarmayor:
@nofir trying to add a one-minute cache on this endpoint. would this work?

### DIFF
--- a/packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
+++ b/packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
@@ -4,7 +4,7 @@ import { Request, RouteOptions } from "@hapi/hapi";
 import _ from "lodash";
 import Joi from "joi";
 import { logger } from "@/common/logger";
-import { fromBuffer, regex, toBuffer } from "@/common/utils";
+import { fromBuffer, regex } from "@/common/utils";
 import { redb } from "@/common/db";
 import * as Sdk from "@reservoir0x/sdk";
 import { config } from "@/config/index";
@@ -24,11 +24,12 @@ const version = "v2";
 
 export const getTopSellingCollectionsV2Options: RouteOptions = {
   cache: {
+    expiresIn: 60 * 1000,
     privacy: "public",
   },
   description: "Top Selling Collections",
   notes: "Get top selling and minting collections",
-  tags: ["api", "Collections"],
+  tags: ["api", "x-deprecated"],
   plugins: {
     "hapi-swagger": {
       order: 3,
@@ -213,13 +214,14 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
       let collections = [];
 
       if (collectionsResult.length) {
-        request.query.contract = collectionsResult.map((collection: any) =>
-          toBuffer(collection.primaryContract)
-        );
+        const collectionIdList = collectionsResult
+          .map((collection: any) => `'${collection.id}'`)
+          .join(", ");
 
         const baseQuery = `
         SELECT
           collections.id,
+          collections.name,
           collections.contract,
           collections.token_count,
           collections.owner_count,
@@ -227,13 +229,14 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
           collections.day7_volume_change,
           collections.day30_volume_change,
           (collections.metadata ->> 'bannerImageUrl')::TEXT AS "banner",
+          (collections.metadata ->> 'imageUrl')::TEXT AS "image",
           (collections.metadata ->> 'description')::TEXT AS "description",
           ${floorAskSelectQuery}
           FROM collections
-          WHERE collections.contract IN ($/contract:csv/)
+          WHERE collections.id IN (${collectionIdList})
       `;
 
-        const resultsPromise = redb.manyOrNone(baseQuery, request.query);
+        const resultsPromise = redb.manyOrNone(baseQuery);
         const recentSalesPromise = collectionsResult.map(async (collection: any) => {
           return {
             ...collection,
@@ -270,14 +273,14 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
         const collectionsMetadata: Record<string, any> = {};
         if (collectionMetadataResponse && Array.isArray(collectionMetadataResponse)) {
           collectionMetadataResponse.forEach((metadata: any) => {
-            collectionsMetadata[fromBuffer(metadata.contract)] = metadata;
+            collectionsMetadata[metadata.id] = metadata;
           });
         }
         const sources = await Sources.getInstance();
 
         collections = await Promise.all(
           responses.map(async (response: any) => {
-            const metadata = collectionsMetadata[(response as any).primaryContract] || {};
+            const metadata = collectionsMetadata[(response as any).id] || {};
             let floorAsk;
             if (metadata) {
               const floorAskCurrency = metadata.floor_sell_currency
@@ -307,8 +310,10 @@ export const getTopSellingCollectionsV2Options: RouteOptions = {
                 "7day": metadata.day7_volume_change,
                 "30day": metadata.day30_volume_change,
               },
+              name: metadata.name,
               tokenCount: Number(metadata.token_count || 0),
               ownerCount: Number(metadata.owner_count || 0),
+              image: metadata.image,
               banner: metadata.banner,
               description: metadata.description,
               floorAsk,
--- a/packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
+++ b/packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
@@ -1,0 +1,335 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { Request, RouteOptions } from "@hapi/hapi";
+import Joi from "joi";
+import { logger } from "@/common/logger";
+import { fromBuffer, regex } from "@/common/utils";
+import { redb } from "@/common/db";
+import * as Sdk from "@reservoir0x/sdk";
+import { config } from "@/config/index";
+import { getStartTime } from "@/models/top-selling-collections/top-selling-collections";
+import { chunk, flatMap } from "lodash";
+
+import { redis } from "@/common/redis";
+
+const REDIS_EXPIRATION = 60 * 60 * 24; // 24 hours
+const REDIS_BATCH_SIZE = 100;
+
+import {
+  getTopSellingCollectionsV2 as getTopSellingCollections,
+  TopSellingFillOptions,
+} from "@/elasticsearch/indexes/activities";
+
+import { getJoiPriceObject, JoiPrice } from "@/common/joi";
+import { Sources } from "@/models/sources";
+
+const version = "v1";
+
+export const getTrendingCollectionsV1Options: RouteOptions = {
+  cache: {
+    expiresIn: 60 * 1000,
+    privacy: "public",
+  },
+  description: "Top Selling Collections",
+  notes: "Get top selling and minting collections",
+  tags: ["api", "Collections"],
+  plugins: {
+    "hapi-swagger": {
+      order: 3,
+    },
+  },
+  validate: {
+    query: Joi.object({
+      period: Joi.string()
+        .valid("5m", "10m", "30m", "1h", "6h", "1d", "24h", "7d", "30d")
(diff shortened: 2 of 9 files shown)


lunarmayor:
things i updated: 
- cache collections after fetching them and keep them up to date with the change events
- increase the limit to 1000 
- improve the speed of and weight of the cache job for the windows
- move the endpoint over to be a dedicated trending collections endpoint
- added cache header for 1m

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,319 @@
...
+        .valid("5m", "10m", "30m", "1h", "6h", "1d", "24h", "7d", "30d")
+        .default("1d")
+        .description("Time window to aggregate."),
+
+      limit: Joi.number()
+        .integer()
+        .min(1)
+        .max(1000)
+        .default(1000)
+        .description("Amount of items returned in response. Default is 25 and max is 50"),
```
lunarmayor, line 51: update to use show max 1000

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,321 @@
...
+        collectionsResult,
+        collectionsMetadata,
+        normalizeRoyalties,
+        useNonFlaggedFloorAsk
+      );
+
+      const response = h.response({ collections });
+      return response;
+    } catch (error) {
+      logger.error(`get-top-selling-collections-${version}-handler`, `Handler failure: ${error}`);
```
pedromcunha, line 138: should we change this so that it reflects the api function name?

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,321 @@
...
+            allTime: Joi.number().unsafe().allow(null),
+          }).description(
+            "Total volume change X-days vs previous X-days. (e.g. 7day [days 1-7] vs 7day prior [days 8-14]). A value over 1 is a positive gain, under 1 is a negative loss. e.g. 1 means no change; 1.1 means 10% increase; 0.9 means 10% decrease."
+          ),
+        })
+      ),
+    }).label(`getTopSellingCollections${version.toUpperCase()}Response`),
+    failAction: (_request, _h, error) => {
+      logger.error(
+        `get-top-selling-collections-${version}-handler`,
```
pedromcunha, line 114: Same here re the name

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,321 @@
...
+  const collectionIds = collectionsResult.map((collection: any) => collection.id);
+  const collectionsToFetch = collectionIds.map((id: string) => `collectionCache:v1:${id}`);
+  const collectionMetadataCache = await redis
+    .mget(collectionsToFetch)
+    .then((results) =>
+      results.filter((result) => !!result).map((result: any) => JSON.parse(result))
+    );
+
+  logger.info(
+    "top-selling-collections",
```
pedromcunha, line 223: name change probably required

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,321 @@
...
+  );
+
+  const collectionsToFetchFromDb = collectionIds.filter((id: string) => {
+    return !collectionMetadataCache.find((cache: any) => cache.id === id);
+  });
+
+  let collectionMetadataResponse: any = [];
+  if (collectionsToFetchFromDb.length > 0) {
+    logger.info(
+      "top-selling-collections",
```
pedromcunha, line 234: name change

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
```diff
@@ -24,6 +24,7 @@ const version = "v2";
 
 export const getTopSellingCollectionsV2Options: RouteOptions = {
   cache: {
+    expiresIn: 60 * 1000,
     privacy: "public",
```
pedromcunha, line 28: we should deprecate this api

packages/indexer/src/jobs/top-selling-collections-cache/index.ts
```diff
@@ -5,29 +5,29 @@ import { acquireLock } from "@/common/redis";
...
 
-// BACKGROUND WORKER ONLY
-if (config.doBackgroundWork) {
-  cron.schedule("*/5 * * * *", async () => {
-    try {
-      if (await acquireLock("top-selling-collections", 10)) {
-        logger.info("top-selling-collections", "Starting top selling collections job");
+// Function to run the job
+async function runJob() {
+  try {
```
pedromcunha, line 10: should we also check if doBackgroundWork is enabled?

packages/indexer/src/models/top-selling-collections/top-selling-collections.ts
```diff
@@ -50,48 +60,34 @@ export const getStartTime = (period: string) => {
...
+    const tasks = fillSorts.flatMap((fillSort) => {
+      return periods.map(async (period) => {
         const startTime = getStartTime(period);
         const topSellingCollections = await getTopSellingCollections({
           startTime,
           fillType: TopSellingFillOptions.sale,
-          limit: 50,
-          includeRecentSales: true,
-          sortBy: "volume",
+          limit: 1000,
```
pedromcunha, line 73: There's a param in the trending api for the limit, is it supposed to be passed into here and used?

pedromcunha (changes requested):

packages/indexer/src/api/endpoints/collections/get-top-selling-collections/v2.ts
lunarmayor: yah we should i think

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
lunarmayor: yah definitely

packages/indexer/src/jobs/top-selling-collections-cache/index.ts
lunarmayor: it gets checked below so I don't think so. basically pulled this out to make it easier to immediately call once

packages/indexer/src/models/top-selling-collections/top-selling-collections.ts
lunarmayor: this is the job so it needs to get the maximum limit that could be returned from the cache

lunarmayor:
@pedromcunha @nofir addressed comments and fixed the conflicts

packages/indexer/src/jobs/cdc/topics/indexer-collections.ts
```diff
@@ -36,6 +39,11 @@ export class IndexerCollectionsHandler extends KafkaEventHandler {
       },
       eventKind: WebsocketEventKind.CollectionEvent,
     });
+    logger.info("top-selling-collections", `updating collection ${payload.after.id}`);
+
+    await redis.set(`collectionCache:v1:${payload.after.id}`, JSON.stringify(payload.after), "XX");
```
nofir, line 44: our naming convention is using dashes vs camelcase

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,324 @@
...
+    })
+  );
+
+  return collections;
+}
+
+async function getCollectionsMetadata(collectionsResult: any[]) {
+  const collectionIds = collectionsResult.map((collection: any) => collection.id);
+  const collectionsToFetch = collectionIds.map((id: string) => `collectionCache:v1:${id}`);
+  const collectionMetadataCache = await redis
```
nofir, line 219: did you test it with 1000 (the previous limit was 50) i worry a bit about blocking redis. is it possible to do in the background job?
or maybe we can lower the limit and introduce pagination

lunarmayor: think saving 1000 is going to happen very infrequently. maybe only the first time. Doesn't seem to be a problem from my testing

packages/indexer/src/jobs/cdc/topics/indexer-collections.ts
lunarmayor: okay can update

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
lunarmayor: reason to return 1000 is ME does client-side filtering and sorting for their trending collections so they need quite a bit to work with. this avoids needing to add a bunch of filtering options for this endpoint

nofir: so there is no plan to support for continuation right now? its always 1000?

1. maybe we can make the max 1000 and the default 50? 
2. you can change the logic to get / save from redis in smaller batches, while still supporting 1000 at the end

we rely on redis less now since we switched to rabbitmq so its less of an issue but its something to be aware of

lunarmayor: yah can do both. For ME we really want them mostly hitting the cdn so it's possible that we would want to make it always 1000. Think it would be fine to set the default to 50 though also

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,335 @@
...
+      return [
+        ["set", `collection-cache:v1:${metadata.id}`, JSON.stringify(metadata)],
+        ["expire", `collection-cache:v1:${metadata.id}`, REDIS_EXPIRATION],
+      ];
+    });
+
+    const commandBatches = chunk(commands, 500);
+
+    for (const batch of commandBatches) {
+      const redisMulti = redis.multi(batch as any);
```
nofir, line 301: so this will make (potentially) 2 blocking multi calls with 1000 commands (set and expire per collection) in each?

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -0,0 +1,335 @@
...
+  let collectionsResult = [];
+  const period = request.query.period === "24h" ? "1d" : request.query.period;
+  const cacheKey = `topSellingCollections:v2:${period}:${fillType}:${sortBy}`;
+  const cachedResults = await redis.get(cacheKey);
+
+  if (cachedResults) {
+    collectionsResult = JSON.parse(cachedResults).slice(0, limit);
+  } else {
+    const startTime = getStartTime(period);
+    collectionsResult = await getTopSellingCollections({
```
nofir, line 326: aren't you missing caching the result once its retrieved from es?

lunarmayor: in this case, it's just using the pre-calculated cache. hopefully people move away from this endpoint in general

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
lunarmayor: it will be 500 commands at a time because of the flat map. so if there is 1000 collections it's going to break into 4 batches of 500 commands (set and expire)

lunarmayor: it's a little hard to say what the right batch size is but figure this this is better than trying to run 2000 commands

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
nofir: but this is the new endpoint - "get-trending-collections/v1" ? im confused...

nofir: So in the new impl, we never cache the result from es?

nofir (approved):

PR #6438, opened 2023-09-25
feat: fill mint with rawMint

lljxx1:


packages/indexer/src/api/endpoints/execute/get-execute-buy/v7.ts
```diff
@@ -92,6 +97,33 @@ export const getExecuteBuyV7Options: RouteOptions = {
               .description(
                 "Optionally specify a particular fill method. Only relevant when filling via `collection`."
               ),
+            rawMint: Joi.object({
```
georgeroman, line 100: Let's use the existing `rawOrder` parameter:
- `kind` should be `mint`
- `data` will be an untyped object, with the fields in the current `rawMint` field

packages/indexer/src/orderbook/mints/calldata/index.ts
```diff
@@ -6,12 +6,24 @@ import { idb } from "@/common/db";
...
 import { CollectionMint } from "@/orderbook/mints";
+import * as Sdk from "@reservoir0x/sdk";
+import { config } from "@/config/index";
 
 import * as mints from "@/orderbook/mints/calldata/detector";
 
 // For now, use the deployer address
 const DEFAULT_REFERRER = "0xf3d63166f0ca56c3c1a3508fce03ff0cf3fb691e";
 
+export type RawMintParam = {
```
georgeroman, line 17: Instead of creating a new type, we can probably use the existing `CollectionMint` type and select certain fields from it using `Pick<CollectionMint, fields>`

packages/indexer/src/orderbook/mints/calldata/index.ts
```diff
@@ -58,6 +70,28 @@ export type MintTxSchema = {
 
 export type CustomInfo = mints.manifold.Info;
 
+export const getCollectionMintFromRawMintParam = (mintRaw: RawMintParam): CollectionMint => {
```
georgeroman, line 73: Let's rename this to `normalizeCollectionMint`

georgeroman (changes requested):

PR #6449, opened 2023-09-25
feat: add custom ens metadata

tv3636:


packages/indexer/src/metadata/custom/ens/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    name: data.name,
+    description: data.description,
+    imageUrl: data.image,
+    imageOriginalUrl: data.image,
+    mediaUrl: data?.background_image,
+    animationOriginalUrl: null,
+    metadataOriginalUrl: url,
+    attributes: data.attributes.map((attribute: any) => {
+      return {
+        key: attribute.trait_type,
```
nofir, line 29: can you provide an example of the response from ens?
what are the possible display types from ens?

nofir (approved):

nofir (approved):

PR #6465, opened 2023-09-26
Tv/platf 2283 return fill source as well as order source in the activity

tv3636:


packages/indexer/src/api/endpoints/activities/get-token-activity/v5.ts
```diff
@@ -294,6 +296,11 @@ export const getTokenActivityV5Options: RouteOptions = {
             : undefined;
         }
 
+        const sources = await Sources.getInstance();
+        const fillSource = activity.event?.fillSourceId
```
nofir, line 300: does it make sense to wait for the pr for the default source? that way you can use the same logic

packages/indexer/src/elasticsearch/indexes/activities/base/index.ts
```diff
@@ -124,6 +125,7 @@ export interface BuildActivityData extends BuildDocumentData {
     data: Record<string, unknown>;
   };
   created_ts: number;
+  fill_source_id?: number;
```
nofir, line 128: can this be event_fill_source_id?

packages/indexer/src/elasticsearch/indexes/activities/base/index.ts
```diff
@@ -174,6 +176,7 @@ export class ActivityBuilder extends DocumentBuilder {
             logIndex: data.event_log_index,
             batchIndex: data.event_batch_index,
             blockHash: fromBuffer(data.event_block_hash!),
+            fillSourceId: data.fill_source_id,
```
nofir, line 179: event_fill_source_id

packages/indexer/src/elasticsearch/indexes/activities/event-handlers/fill-event-created/index.ts
```diff
@@ -71,6 +71,7 @@ export class FillEventCreatedEventHandler extends BaseActivityEventHandler {
                   order_kind,
                   order_side,
                   order_source_id_int,
+                  fill_source_id,
```
nofir, line 74: event_fill_source_id

nofir (changes requested):

packages/indexer/src/api/endpoints/activities/get-token-activity/v5.ts
tv3636: don't we actually not want to use it here since we'd have `fillSource: null` on non-sale events?

nofir (approved):

nofir (approved):

PR #6489, opened 2023-09-28
fix: update log to debug onchain metadata errors

tv3636:


packages/indexer/src/metadata/providers/onchain-metadata-provider.ts
```diff
@@ -442,7 +446,10 @@ export class OnchainMetadataProvider extends AbstractBaseMetadataProvider {
       const json = await response.json();
       return [json, null];
     } catch (e) {
-      logger.error("onchain-fetcher", `getTokenMetadataFromURI error. error:${e}`);
+      logger.error(
```
nofir, line 449: i would suggest making this a json obj and add the error as a property, really helps with debugging

nofir (dismissed):

nofir (approved):

PR #6492, opened 2023-09-28
feat: add metadata admin API

tv3636:


packages/indexer/src/api/endpoints/admin/get-metadata.ts
```diff
@@ -0,0 +1,82 @@
...
+      query.tokens = [query.tokens];
+    }
+
+    try {
+      if (params.type === "collection") {
+        const [contract, tokenId] = query.tokens[0].split(":");
+        return await MetadataProviderRouter.getCollectionMetadata(
+          contract,
+          tokenId,
+          query.community ?? "",
```
nofir, line 63: i think this should not be passed from the client. either the method was passed or we get based on the logic we have (community /env var)

packages/indexer/src/api/routes.ts
```diff
@@ -184,6 +184,12 @@ export const setupRoutes = (server: Server) => {
     options: adminEndpoints.getRateLimitRulesOptions,
   });
 
+  server.route({
+    method: "GET",
+    path: "/admin/metadata/{type}",
```
nofir, line 189: i think this should be more explicit - get provider metadata/ get metadata from provider / something like that (also in the file names)

nofir (changes requested):

nofir (approved):

PR #6531, opened 2023-10-04
feat: allow filter on fill/order source for sales ws

tv3636:


packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
```diff
@@ -143,6 +143,8 @@ export class SaleWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJobHandl
           contract: data.after.contract,
           maker: data.after.maker,
           taker: data.after.taker,
+          fillSource: result.fillSource ?? "",
```
nofir, line 146: Isnt the fillSource / orderSource the whole obj? it should be based on the domain

nofir (changes requested):

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
tv3636: `getJoiSaleObject` returns just the domain: https://github.com/reservoirprotocol/indexer/blob/8dd3a14d4e16bdd0fb254e48b17b23ad928decae/packages/indexer/src/common/joi.ts#L909

nofir: i missed that, ok makes sense. any reason to set it to ""? just trying to understand if not including it (undefined), creates an issues?

tv3636: the ws publisher expects all tags to be `string`, alternatively we could allow the tags to be null and remove the empty string here

nofir: and again, if the firehose, just ignores empty strings, its probably ok to keep like that, i just thought it was cleaner

nofir (approved):

PR #6532, opened 2023-10-04
feat: support token name search by id

tv3636:


packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -784,8 +784,13 @@ export const getTokensV6Options: RouteOptions = {
       }
 
       if (query.tokenName) {
+        if (!isNaN(query.tokenName)) {
```
nofir, line 787: not sure this was the intention. basically if in the db the token name is missing , search the token id , with whatever was passed from the client. see bored ape for example (no token name)

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
tv3636: so we'd have 2 queries then? i think it makes sense to do it at once since it's pretty common for token names to have the tokenId and the response will be faster

nofir: but the tokenName being passed from the client is not going to be null. today if you pass "1" when you search for bored apes, you would not get results, right? (because the name in the db is null)

wouldn't you be able to do:
```conditions.push(`COALESCE (t.name,t.token_id) ILIKE $/tokenName/`);```

maybe im missing something

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
tv3636: `!isNaN(query.tokenName)` evaluates to true if the string is a valid number (and we already checked above for `query.tokenName` not being null, so in cases where the search string is a number, we would also check the tokenId for matches

nofir: if you're going with this approach of checking both, i would change the token id compare to equal. i dont think we want to do a LIKE , you dont want to get 111 token if someone is looking for 1

nofir (changes requested):

nofir (approved):

PR #6545, opened 2023-10-05
feat: fix 429 metadata

d3or:


packages/indexer/src/metadata/providers/simplehash-metadata-provider.ts
```diff
@@ -90,6 +78,24 @@ export class SimplehashMetadataProvider extends AbstractBaseMetadataProvider {
     throw new Error("Method not implemented.");
   }
 
+  handleError(error: any) {
+    if (error.response?.status === 429 || error.response?.status === 503) {
```
nofir, line 82: he 503 is only relevant for opensea

nofir (approved):
looks good, left one comment but no a real issue

PR #6615, opened 2023-10-13
Add trending collections endpoint

lunarmayor:
@nofir opening back up. Had to fix the old endpoints when fetching recent sales. Think very likely we'll just remove the v1 of the top-selling soon anyway. Also added the percent change which is something that Magic Eden really needs

### DIFF
--- a/packages/indexer/src/jobs/top-selling-collections-cache/index.ts
+++ b/packages/indexer/src/jobs/top-selling-collections-cache/index.ts
@@ -5,29 +5,29 @@ import { acquireLock } from "@/common/redis";
 import { config } from "@/config/index";
 import { topSellingCollectionsJob } from "@/jobs/top-selling-collections-cache/save-top-selling-collections-job";
 
+// Function to run the job
+async function runJob() {
+  try {
+    if (await acquireLock("top-selling-collections", 10)) {
+      logger.info("top-selling-collections", "Starting top selling collections job");
+
+      try {
+        await topSellingCollectionsJob.addToQueue();
+      } catch (error) {
+        logger.error("top-selling-collections", `Failed to save top selling collections: ${error}`);
+      }
+    }
+  } catch (error) {
+    logger.error(
+      "top-selling-collections",
+      JSON.stringify({
+        msg: error,
+      })
+    );
+  }
+}
+
 // BACKGROUND WORKER ONLY
 if (config.doBackgroundWork) {
-  cron.schedule("*/5 * * * *", async () => {
-    try {
-      if (await acquireLock("top-selling-collections", 10)) {
-        logger.info("top-selling-collections", "Starting top selling collections job");
-
-        try {
-          await topSellingCollectionsJob.addToQueue();
-        } catch (error) {
-          logger.error(
-            "top-selling-collections",
-            `Failed to save top selling collections: ${error}`
-          );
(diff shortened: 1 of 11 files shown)


lunarmayor:
testing more thoroughly on dev right now

packages/indexer/src/jobs/top-selling-collections-cache/index.ts
```diff
@@ -5,29 +5,29 @@ import { acquireLock } from "@/common/redis";
...
+      JSON.stringify({
+        msg: error,
+      })
+    );
+  }
+}
+
+// BACKGROUND WORKER ONLY
+if (config.doBackgroundWork) {
+  cron.schedule("*/10 * * * *", runJob);
```
nofir, line 32: you need to take a lock, otherwise, each running pod of the service will trigger the job.
```  cron.schedule(
    "*/5 * * * * *",
    async () =>
      await redlock
        .acquire([`${tokenFlagStatusSyncJob.queueName}-cron-lock`], (5 - 1) * 1000)
        .then(async () => tokenFlagStatusSyncJob.addToQueue())
        .catch(() => {
          // Skip on any errors
        })
  );```

Comment (nofir) [reply packages/indexer/src/jobs/top-selling-collections-cache/index.ts:32]: oh i see you are doing inside the job, so up to you, you can simplify it with redlock

File: packages/indexer/src/models/top-selling-collections/top-selling-collections.ts [hunk 2]
```diff
@@ -50,48 +60,34 @@ export const getStartTime = (period: string) => {
...
+    const tasks = fillSorts.flatMap((fillSort) => {
+      return periods.map(async (period) => {
         const startTime = getStartTime(period);
         const topSellingCollections = await getTopSellingCollections({
           startTime,
           fillType: TopSellingFillOptions.sale,
-          limit: 50,
-          includeRecentSales: true,
-          sortBy: "volume",
+          limit: 1000,
```
nofir, line 73: So we always only cache the top 1000? and then the pagination is done on the endpoint?

packages/indexer/src/jobs/top-selling-collections-cache/save-top-selling-collections-job.ts
```diff
@@ -6,6 +6,7 @@ export type TopSellingCollectionsJobPayload = {
   retry: number;
 };
 
+TopSellingCollections.updateTopSellingCollections();
```
nofir, line 9: Why is this being called here?

packages/indexer/src/models/top-selling-collections/top-selling-collections.ts
lunarmayor: yah just cache the max that will get returned

packages/indexer/src/jobs/top-selling-collections-cache/index.ts
lunarmayor: ah interesting. didn't know that

lunarmayor: probably will keep it the same for right now

lunarmayor:
@nofir good to go?

nofir (approved):

PR #6626, opened 2023-10-16
feat: mintdotfun

lljxx1:


georgeroman:
A few comments:
- I assume the mint interface is different from `mintdotfun` we already handle, so let's rename it to `createfun`
- let's resolve the conflicts, you're probably branching out of an old `main` version

PR #6644, opened 2023-10-16
Add val-erc-20

Nike682631:


packages/indexer/src/config/network.ts
```diff
@@ -439,6 +439,19 @@ export const getNetworkSettings = (): NetworkSettings => {
               },
             },
           ],
+          [
+            "0x011e128ec62840186f4a07e85e3ace28858c5606",
+            {
+              contract: "0x011e128ec62840186f4a07e85e3ace28858c5606",
+              name: "Val",
+              symbol: "$VAL",
+              decimals: 2,
```
nofir, line 448: Is there a reason this is 2 and not 18?

nofir (approved):

PR #6663, opened 2023-10-17
feat: extend kanpai pandas metadata

jasonyjong:


packages/indexer/src/metadata/extend/kanpai-pandas/index.ts
```diff
@@ -0,0 +1,29 @@
+import axios from "axios";
+import { TokenMetadata } from "@/metadata/types";
+
+const metadataBaseURI = "https://metadata-api-snag-solutions-ad3b2d55c570.herokuapp.com/api/token";
```
nofir, line 4: @jasonyjong The url does not seem to work from my tests. (times out)

jasonyjong: Hmm works for me now -- I wonder if it's cause heroku spins down their instance. Will message them Kanpai team.

nofir (approved):

packages/indexer/src/metadata/extend/kanpai-pandas/index.ts
nofir: let me know if I should wait with merging it incase you need to make a change.

jasonyjong: We should be good to merge -- will just have their team look into it. It should be up for at least 24 hours, but that's only if 0 requests are sent in that time period

PR #6668, opened 2023-10-17
feat: seaport private listing

lljxx1:


georgeroman:
@lljxx1 We should also support X2Y2

PR #6683, opened 2023-10-17
fix: emit less logs for updating collections

lunarmayor:
maybe should just remove these in general now


lunarmayor:
just emitting too many logs so changed it to only log when a collection actually needs to be updated

PR #6684, opened 2023-10-17
Val-polygon

Nike682631:
Add val token to polygon


packages/indexer/src/config/network.ts
```diff
@@ -701,6 +701,19 @@ export const getNetworkSettings = (): NetworkSettings => {
...
+          [
+            "0x011e128ec62840186f4a07e85e3ace28858c5606",
+            {
+              contract: "0x011e128ec62840186f4a07e85e3ace28858c5606",
+              name: "WVAL",
+              symbol: "$VAL",
+              decimals: 18,
+              metadata: {
+                image:
+                  "https:/ /i.ibb.co/s1k4Qvz/valeria-logo.png",
```
jasonyjong, line 713: @Nike682631 there's an extra space in there

tv3636 (approved):

PR #6685, opened 2023-10-17
feat: remove logs

d3or:


lunarmayor (approved):
thanks! meant to remove these logs after i verified it was working but seems to be putting out a lot 😅

PR #6693, opened 2023-10-18
Armando/grwth 3385 trending mints api endpoint org

r3lays:


packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,533 @@
...
+          ),
+        })
+      ),
+    }).label(`get-trending-mints${version.toUpperCase()}Response`),
+    failAction: (_request, _h, error) => {
+      logger.error(`get-trending-mints-${version}-handler`, `Wrong response schema: ${error}`);
+      throw error;
+    },
+  },
+  handler: async ({ query }: Request, h) => {
```
lunarmayor, line 219: I feel like there are a couple of issues there. the big one I'm seeing is that the `getMintingCollections` is getting limited which means that when you're aggregating you're shrinking the amount of collections that you are looking at which removes a lot of valid results. Really need to be returning the whole set of possible minting collections and then aggregate those to get the trending set.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,533 @@
...
+};
+
+async function getElasticMints(
+  contracts: string[],
+  startTime: number
+): Promise<ElasticMintResult[]> {
+  return getTrendingMintsV2({ contracts, startTime });
+}
+
+async function getMintingCollections(
```
lunarmayor, line 254: think this can get simplified. feel like the goal of this function should be to return the set of minting collections which means you can avoid getting the mint stages here and also constrain this to only public mints

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  const combinedResults = cachedResults.concat(resultsFromDB);
+
+  if (resultsFromDB.length > 0) {
+    await redis.set(cacheKey, JSON.stringify(combinedResults));
+    await redis.expire(cacheKey, REDIS_EXPIRATION_MINTS);
+  }
+
+  return combinedResults.slice(0, limit);
```
lunarmayor, line 321: should return the total set of minting collections. this just cuts off the possible aggregation to just the most recent collections. If you want the top 50 this is saying what are the 50 most recent mints and of those what have had the most mints

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,533 @@
...
+      collections.normalized_floor_sell_id,
+      collections.normalized_floor_sell_value,
+      collections.normalized_floor_sell_maker,
+      collections.normalized_floor_sell_valid_between,
+      collections.normalized_floor_sell_source_id_int,
+      collections.top_buy_id,
+      collections.top_buy_value,
+      collections.top_buy_maker,
+      collections.top_buy_valid_between,
+      collections.top_buy_source_id_int
```
lunarmayor, line 480: this is really where mint stages should be fetched i think. since you really only need mint stages for the collections that are getting returned.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,533 @@
...
+  logger.info(
+    "top-selling-collections",
+    `using ${collectionMetadataCache.length} collections from cache`
+  );
+
+  const collectionsToFetchFromDb = collectionIds.filter((id: string) => {
+    return !collectionMetadataCache.find((cache: any) => cache.id === id);
+  });
+
+  let collectionMetadataResponse: any = [];
```
lunarmayor, line 435: to start i think we can remove all the caching from this. If it helps we can limit the number of collections to 25 and should be able to fetch those every time. Just going to be a lot more work to keep this cache up to date

lunarmayor (commented):
Left some more specific comments but also sketched out a possible plan that we can chat about when you get back 

https://jasonandmichael.invisionapp.com/freehand/Trending-Mints-qKQjJPtnZ

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -604,6 +605,90 @@ export const getTopSellingCollectionsV2 = async (params: {
...
+            },
+          },
+        },
+        {
+          terms: {
+            "collection.id": contracts,
+          },
+        },
+      ],
+      ...(trendingExcludedContracts && {
```
lunarmayor, line 638: don't think we need this right now

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -604,6 +605,90 @@ export const getTopSellingCollectionsV2 = async (params: {
...
+        ],
+      }),
+    },
+  } as any;
+
+  const collectionAggregation = {
+    collections: {
+      terms: {
+        field: "collection.id",
+        size: 50,
```
lunarmayor, line 654: should be the limit here

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: Removed the limit and now we're returning the entire table ~11k avg. Only concern is it might take slightly longer to query. Are we supposed to cache these?

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: Resolved. I'm still returning the mint_stages however it's only public mints now where the status is open.
Why wouldnt we return the mint_stages? If we're already querying here it doesn't make sense to do it again on the metadata query.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: Resolved.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: ah I see your logic. 
I'll test out the performance between the two but I believe that getting them from collection_mints might be faster. Since we only return the collection_id and array_agg the mint stages.

PR #6705, opened 2023-10-19
feat: add typeorm

lljxx1:
## File structure

```shell
src/data-source.ts
src/migration
src/entities
```

## Init migration
run migrate but actually doesn't execute any SQL, used to set the database state to the latest migration 

```shell
yarn typeorm-migrate-init
```

## Migration
```shell
yarn typeorm-migrate

# revert to prev step
yarn typeorm-revert
```

## Codebase Sync
auto sync Migration or Entitiy definitions  to the `src/data-source.ts`
```
yarn typeorm-sync
```


georgeroman:
Nice, a few questions / notes:
- How does the new migration logic handles the old migrations (which I see were ported over)? Ideally it will skip them and only pick up the new ones (we don't want to have the old migrations re-run since that could create issues).
- We should probably remove the `node-pg-migrate` dependency if it's not needed anymore together with any scripts for it (`yarn migrate`).

cc @nofir

lljxx1:
> Nice, a few questions / notes:
> 
> * How does the new migration logic handles the old migrations (which I see were ported over)? Ideally it will skip them and only pick up the new ones (we don't want to have the old migrations re-run since that could create issues).
> * We should probably remove the `node-pg-migrate` dependency if it's not needed anymore together with any scripts for it (`yarn migrate`).
> 
> cc @nofir

all the old migrations were converted to the TypeORM migration style, and the migration state will be stored in the `typeorm_migrations` table.

run `yarn typeorm-migrate-init` to skip all

packages/indexer/package.json
```diff
@@ -14,7 +14,11 @@
     "debug": "yarn build && yarn migrate up; node --inspect -r module-alias/register ./dist/index.js",
```
nofir, line 14: i assume that once you do the initial init, you would change this to typeorm-migrate

PR #6756, opened 2023-10-25
feat: websocket events test

lljxx1:


packages/indexer/package.json
```diff
@@ -14,7 +14,8 @@
     "debug": "yarn build && yarn migrate up; node --inspect -r module-alias/register ./dist/index.js",
     "format": "prettier --write \"src/**/*.+(js|ts|json)\"",
     "cli": "node -r module-alias/register ./dist/cli/index.js",
-    "test": "jest --no-watchman"
+    "test": "jest --no-watchman",
+    "ci-test": "jest --no-watchman src/tests/websocket/**.test.ts"
```
nofir, line 18: is that just for local dev?

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
```diff
@@ -33,162 +33,152 @@ export class SaleWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJobHandl
     delay: 1000,
   } as BackoffStrategy;
 
-  protected async process(payload: SaleWebsocketEventsTriggerQueueJobPayload) {
+  static async format(payload: SaleWebsocketEventsTriggerQueueJobPayload) {
```
nofir, line 36: Just to i can understand, what is the reasoning for needing that?

can't you initiate a new instance of SaleWebsocketEventsTriggerQueueJob and call the process when you do the test?

packages/indexer/src/tests/websocket/sale.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    //     if (typeof value === "string" && value.includes("=")) {
+    //       (payload as any)[key] = fromBuffer(Buffer.from(value, "base64"));
+    //     } else if (value !== null && typeof value === "object") {
+    //       fixBase64(payload[key]);
+    //     }
+    //   }
+    // }
+    // fixBase64(payload);
+
+    const message = await SaleWebsocketEventsTriggerQueueJob.format({
```
nofir, line 57: following my question about the need for format, wouldn't it be possible to assert `toBeCalledWith` on publishWebsocketEvent?

I think it has the same outcome but this approach looks what goes in and out of the process.

packages/indexer/src/tests/websocket/sale.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      data: payload,
+    } as unknown as SaleWebsocketEventsTriggerQueueJobPayload);
+
+    const status = JoiSale.validate(message!.data);
+
+    expect(status.error).toBe(undefined);
+    expect(message!.event).toBe("sale.created");
+    expect(message!.tags.contract).toBe("0x4e76c23fe2a4e37b5e07b5625e17098baab86c18");
+    expect(message!.tags.maker).toBe("0xd805cec34482ee455144dd04aad09f2758cddfe8");
+    expect(message!.tags.taker).toBe("0x98415e65216c83910f82f663dc152a535c0428b5");
```
nofir, line 67: I assume this is incomplete

packages/indexer/package.json
lljxx1: we could use this to run the tests in CI

packages/indexer/src/jobs/websocket-events/sale-websocket-events-trigger-job.ts
lljxx1: addressed, but `process` is protected, it's can't be called outside. 
so I added `handle(payload: any)` to AbstractRabbitMqJob Handler to process the payload synchronized

nofir (approved):
looks good. left a comment with a question

Also, because of the change in code, lets just make sure we merge it to development and angel can confirm there is not issue with sale ws events

packages/indexer/src/jobs/abstract-rabbit-mq-job-handler.ts
```diff
@@ -174,6 +174,10 @@ export abstract class AbstractRabbitMqJobHandler {
     }
   }
 
+  public async handle(payload: any) {
```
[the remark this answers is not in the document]
lljxx1: @ipeleg OK!

PR #6770, opened 2023-10-26
feat: filter spam from trending collections

lunarmayor:


packages/indexer/src/models/top-selling-collections/top-selling-collections.ts
```diff
@@ -59,10 +61,26 @@ export const getStartTime = (period: Period): number => {
...
+      collections.id
+    FROM collections
+    WHERE
+      day30_volume > 0 AND
+      is_spam = 1
+  `;
+
+  const results = await redb.manyOrNone(query);
+  const ids = results.map((r) => r.id);
+  await redis.set("active-spam-collection-ids", JSON.stringify(ids));
```
nofir, line 76: i would just maybe add a try and catch here to make sure that it does not fail updating the collections entirely

nofir (approved):

PR #6771, opened 2023-10-26
feat: added wash trading score

nofir:


lunarmayor (approved):
looks good to me. do you still need that debugging log?

PR #6806, opened 2023-10-30
feat: support spam report api

ipeleg:


packages/indexer/src/api/endpoints/management/post-report-spam/v1.ts
```diff
@@ -0,0 +1,155 @@
...
+import { idb } from "@/common/db";
+import {
+  ActionsContext,
+  ActionsOrigin,
+  actionsTrackingJob,
+} from "@/jobs/actions-tracking/actions-tracking-job";
+
+const version = "v1";
+
+export const postSimulateOrderV1Options: RouteOptions = {
```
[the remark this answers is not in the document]
nofir: also, what about adding it to the routes?

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1289,7 +1289,7 @@ export const getTokensV6Options: RouteOptions = {
             media: r.media,
             kind: r.kind,
             isFlagged: Boolean(Number(r.is_flagged)),
-            isSpam: Boolean(Number(r.t_is_spam)) || Boolean(Number(r.c_is_spam)),
+            isSpam: Number(r.t_is_spam) > 0 || Number(r.c_is_spam) > 0,
```
nofir, line 1292: we really need to make the token response generic using joi (similar to order, sales, etc)...

packages/indexer/src/jobs/actions-tracking/actions-tracking-job.ts
```diff
@@ -0,0 +1,61 @@
...
+import { idb } from "@/common/db";
+
+export enum ActionsOrigin {
+  DailyProcess = "daily-process",
+  CollectionRefresh = "collection-refresh",
+  API = "api",
+}
+
+export enum ActionsContext {
+  SpamContractUpdate = "spam-contract-update",
```
nofir, line 11: i would change it to "spam score changed" or something like that, given that peter mentioned that is more likely to be something that can vary. we can discuss with him

packages/indexer/src/jobs/actions-tracking/actions-tracking-job.ts
```diff
@@ -0,0 +1,61 @@
...
+  SpamCollectionUpdate = "spam-collection-update",
+  SpamTokenUpdate = "spam-token-update",
+  DisableMetadataUpdate = "disable-metadata-update",
```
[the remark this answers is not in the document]
nofir: I would change this to be more like the "events" tables (maybe one for contract, token, collection?)

packages/indexer/src/jobs/collections-refresh/collections-refresh-spam-job.ts
```diff
@@ -37,18 +42,31 @@ export default class CollectionRefreshSpamJob extends AbstractRabbitMqJobHandler
...
-          AND is_spam IS DISTINCT FROM 1
+          AND (is_spam IS NULL OR is_spam = 0)
         `;
 
         await idb.none(query, {
           newSpamContracts: newSpamContracts.map(toBuffer),
         });
+
+        // Track the change
+        const trackingParams = _.map(newSpamContracts, (contract) => ({
```
nofir, line 59: any reason to do a job and not do it inside the query? similar to how we do events

packages/indexer/src/jobs/collections-refresh/collections-refresh-spam-job.ts
```diff
@@ -37,18 +42,31 @@ export default class CollectionRefreshSpamJob extends AbstractRabbitMqJobHandler
...
 
-      // Update the new spam collections
+      // Update the new spam collections but don't overwrite spam settings done by users
       if (!_.isEmpty(newSpamContracts)) {
         const query = `
           UPDATE collections
           SET is_spam = 1, updated_at = now()
           WHERE contract IN ($/newSpamContracts:list/)
-          AND is_spam IS DISTINCT FROM 1
+          AND (is_spam IS NULL OR is_spam = 0)
```
nofir, line 51: We dont have an opposite case? where a contract is no longer marked as spam by alchemy?

packages/indexer/src/models/collections/index.ts
```diff
@@ -202,8 +207,21 @@ export class Collections {
     }
 
     const isSpamContract = await AlchemyApi.isSpamContract(collection.contract);
-    if (isSpamContract) {
```
nofir, line 205: why not check first if the collection is already marked as spam on the collections table?

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
ipeleg: This is something Tom was working on, probably should be separate story but he did some work there already

packages/indexer/src/jobs/collections-refresh/collections-refresh-spam-job.ts
ipeleg: Yes, job would be safer and will take less time so it won't hold any request or make any process longer than what it is

packages/indexer/src/jobs/collections-refresh/collections-refresh-spam-job.ts
ipeleg: We don't support that, I wonder if it's even an option

packages/indexer/src/jobs/actions-tracking/actions-tracking-job.ts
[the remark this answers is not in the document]
ipeleg: Yeah, let's discuss this events / actions is kinda the same, don't mind changing it, I actually do this things should be tracked in one place as it's mostly for internal use for us to better investigate issues, don't see a reason to have a dedicated table for each "event" we want to track, feels like it would defeat the purpose in a way

packages/indexer/src/jobs/actions-tracking/actions-tracking-job.ts
ipeleg: I feel like it same meaning, spam score changed / spam contract update 😅

packages/indexer/src/api/routes.ts
```diff
@@ -1173,6 +1179,12 @@ export const setupRoutes = (server: Server) => {
 
   // Tokens
 
+  server.route({
+    method: "POST",
+    path: "/tokens/{token}/report-spam/v1",
```
nofir, line 1184: i think we need to consistent with user facing apis and not have "report-spam", unless we do it somewhere else outside of admin apis... maybe it can be "spam-status"

ipeleg: we don't do it all today, but what do you mean by consistent? consistent with what?

nofir: being restful, we don't put commands in the route

PR #6827, opened 2023-10-31
feat: trending mints ep

r3lays:


packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,497 @@
...
+  const conditions: string[] = [];
+  conditions.push(`kind = 'public'`, `status = 'open'`);
+  type && type !== "any" && conditions.push(`price ${type === "free" ? "= 0" : "> 0"}`);
+
+  const whereClause = conditions.length ? `WHERE ${conditions.join(" AND ")}` : "";
+
+  const baseQuery = `
+SELECT 
+    collection_id,
+    array_agg(
```
lunarmayor, line 268: still feel like this is the wrong place to fetch this data. won't it mostly be thrown away?  you only need the mint stages for the final 50. maybe this is still more efficient even with that said. wdyt?

lunarmayor: guess this does feel simpler for right now. alternative is the last collection fetch but that's getting cached

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,497 @@
...
+        ownerCount: Number(metadata.owner_count || 0),
+        floorAsk,
+      };
+    })
+  );
+
+  return collections;
+}
+
+async function getCollectionsMetadata(collectionIds: string[]): Promise<Record<string, Metadata>> {
```
lunarmayor, line 388: wonder if should try to combine the trending collection function for this and this function. wouldn't it just be the same?

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -343,6 +344,80 @@ export const getTopSellingCollections = async (params: {
...
+  contracts: string[];
+  startTime: number;
+  limit: number;
+}): Promise<ElasticMintResult[]> => {
+  const { contracts, startTime, limit } = params;
+
+  const salesQuery = {
+    bool: {
+      filter: [
+        {
```
lunarmayor, line 357: think we can use term here if it's just the one

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -0,0 +1,497 @@
...
+            })
+              .allow(null)
+              .description("Lowest Ask Price."),
+          },
+          mintPrice: Joi.number().allow(null),
+          mintVolume: Joi.any(),
+          mintCount: Joi.number().allow(null),
+          mintType: Joi.string().allow("free", "paid", "", null),
+          mintStatus: Joi.string().allow("", null),
+          mintStages: Joi.array().items(
```
lunarmayor, line 189: wonder if we should pull out more of the public mint stage like end and start, end, maxMints.

r3lays: Yeah we can do that.

packages/indexer/src/elasticsearch/indexes/activities/index.ts
r3lays: sg will fix.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: Ill test the performance out to determine if we need the changes.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: In the trending collections we dont return the mint_stages.

PR #6833, opened 2023-11-01
Removing $VAL eth version and adding METH token

Nike682631:
Adding METH token for BAP Methane and removing VAL eth token so the coingecko one could be accessed.

### DIFF
--- a/packages/indexer/src/config/network.ts
+++ b/packages/indexer/src/config/network.ts
@@ -438,14 +438,14 @@ export const getNetworkSettings = (): NetworkSettings => {
             },
           ],
           [
-            "0x011e128ec62840186f4a07e85e3ace28858c5606",
+            "0xed5464bd5c477b7f71739ce1d741b43e932b97b0",
             {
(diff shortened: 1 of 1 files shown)


packages/indexer/src/config/network.ts
```diff
@@ -438,14 +438,14 @@ export const getNetworkSettings = (): NetworkSettings => {
...
-            "0x011e128ec62840186f4a07e85e3ace28858c5606",
+            "0xed5464bd5c477b7f71739ce1d741b43e932b97b0",
             {
-              contract: "0x011e128ec62840186f4a07e85e3ace28858c5606",
-              name: "Val",
-              symbol: "$VAL",
+              contract: "0xed5464bd5c477b7f71739ce1d741b43e932b97b0",
+              name: "BAP Methane",
+              symbol: "METH",
               decimals: 18,
```
nofir, line 446: should the decimals still be 18 for METH?

ipeleg (approved):

PR #6835, opened 2023-11-01
Scale up onchain indexing

d3or:


packages/indexer/src/jobs/metadata-index/metadata-fetch-job.ts
```diff
@@ -140,10 +142,15 @@ export default class MetadataIndexFetchJob extends AbstractRabbitMqJobHandler {
     }
 
     // Add the tokens to the list
-    const pendingRefreshTokens = new PendingRefreshTokens(data.method);
-    await pendingRefreshTokens.add(refreshTokens, prioritized);
-
-    await metadataIndexProcessJob.addToQueue({ method: data.method });
+    if (data.method === "onchain") {
+      await PendingFetchOnchainUriTokens.add(refreshTokens, prioritized);
```
nofir, line 146: why do we need a different redis queue? its already per method

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-uri-job.ts
```diff
@@ -0,0 +1,92 @@
...
+  queueName = "metadata-index-onchain-process-uri-queue";
+  maxRetries = 3;
+  concurrency = 3;
+  timeout = 5 * 60 * 1000;
+  backoff = {
+    type: "exponential",
+    delay: 20000,
+  } as BackoffStrategy;
+
+  protected async process() {
```
nofir, line 21: should this job be singleActiveConsumer? only max 50 in any given time?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-uri-job.ts
```diff
@@ -0,0 +1,92 @@
...
+    // If there are potentially more token uris to process, trigger another job
+    const queueLength = await PendingFetchOnchainUriTokens.len();
+    if (queueLength > 0) {
+      return 1;
+    }
+
+    return 0;
+  }
+
+  public async onCompleted(rabbitMqMessage: RabbitMQMessage, processResult: undefined | number) {
```
nofir, line 81: any reason you are using the onCompleted? in line 43 you are just calling addToQueue

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-uri-job.ts
```diff
@@ -0,0 +1,92 @@
...
+import _ from "lodash";
+
+import { logger } from "@/common/logger";
+import { RabbitMQMessage } from "@/common/rabbit-mq";
+import { onchainMetadataProvider } from "@/metadata/providers/onchain-metadata-provider";
+import { onchainMetadataIndexProcessJob } from "./onchain-metadata-process-job";
+import { RequestWasThrottledError } from "@/metadata/providers/utils";
+import { PendingFetchOnchainUriTokens } from "@/models/pending-fetch-onchain-uri-tokens";
+
+export default class OnchainMetadataProcessUriJob extends AbstractRabbitMqJobHandler {
```
nofir, line 11: can this job name be get OnchainMetadataFetchTokenUriJob?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-job.ts
```diff
@@ -0,0 +1,71 @@
...
+import { onchainMetadataProvider } from "@/metadata/providers/onchain-metadata-provider";
+import { RequestWasThrottledError } from "@/metadata/providers/utils";
+
+export type OnchainMetadataIndexProcessJobPayload = {
+  contract: string;
+  tokenId: string;
+  uri: string;
+};
+
+export default class OnchainMetadataIndexProcessJob extends AbstractRabbitMqJobHandler {
```
nofir, line 13: can this job name be get OnchainMetadataProcessTokenUriJob?

nofir (changes requested):

packages/indexer/src/api/endpoints/admin/post-update-image-version.ts
```diff
@@ -0,0 +1,73 @@
...
+        );
+        return { message: "Request accepted" };
+      }
+
+      const collectionId = payload.collection;
+
+      // update all tokens in the collection
+      await idb.oneOrNone(
+        `
+        UPDATE tokens
```
nofir, line 61: that might be risky with very big collections. i guess it will just timeout. I would add seperate logic to do that in batching. (not urgent for now, either remove it or we can make sure people know to be careful with it)

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-job.ts
```diff
@@ -0,0 +1,85 @@
...
+        return;
+      }
+      logger.error(
+        this.queueName,
+        `Error. contract=${contract}, tokenId=${tokenId}, uri=${uri}, error=${e}`
+      );
+    }
+
+    // for whatever reason, we didn't find the metadata, we fallback to simplehash
+    const pendingRefreshTokens = new PendingRefreshTokens("simplehash");
```
nofir, line 60: does it make sense to add a log here to see how often this happens?
also, there are chains that simplehash does not support so we might need to have a FALLBACK_INDEXING_METHOD=simplehash env var and only do if its set

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
```diff
@@ -0,0 +1,104 @@
...
+          contract: result.contract,
+          tokenId: result.tokenId,
+        });
+      }
+    });
+
+    await onchainMetadataProcessTokenUriJob.addToQueueBulk(tokensToProcess);
+
+    // Default to simple hash if no uri found
+    if (!_.isEmpty(fallbackTokens)) {
```
nofir, line 86: same comment as above

nofir: also, did we discuss just getting the metdata from simplehash and then passing the token uri from the simplehash response to the process job?

packages/indexer/src/migrations/1699888093189_tokens_image_refreshed_at.sql
```diff
@@ -0,0 +1,7 @@
+-- Up Migration
+ALTER TABLE "tokens" ADD COLUMN "image_version_updated_at" TIMESTAMPTZ;
```
nofir, line 2: we discussed doing the migration in a seperate pr in order to avoid any issues with failed migration / code migration (until we have the logic to fail deployment upon migration fail)

packages/indexer/src/jobs/metadata-index/metadata-fetch-job.ts
```diff
@@ -139,11 +142,27 @@ export default class MetadataIndexFetchJob extends AbstractRabbitMqJobHandler {
...
 
+    const fetchTokensOnchain = refreshTokens.filter(
+      (token) => !hasExtendCollectionHandler(token.contract) && data.method === "onchain"
+    );
+
     // Add the tokens to the list
-    const pendingRefreshTokens = new PendingRefreshTokens(data.method);
-    await pendingRefreshTokens.add(refreshTokens, prioritized);
+    if (fetchTokensOnchain.length > 0) {
+      await PendingFetchOnchainUriTokens.add(fetchTokensOnchain, prioritized);
```
nofir, line 151: can you remind me again why we need PendingFetchOnchainUriTokens? why not using PendingRefreshTokens("onchain")?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-job.ts
d3or: there is a warn log above that logs when the URI is not found.

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
d3or: i dont think we discussed that

packages/indexer/src/api/endpoints/admin/post-update-image-version.ts
d3or: i added a limit to 10k tokens to be updated at once for now

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
nofir: Maybe we can just add a log here to see how often it happens. I think this is ok for now but the goal is to not have simplehash / opensea images

packages/indexer/src/jobs/metadata-index/metadata-write-job.ts
```diff
@@ -85,19 +85,6 @@ export default class MetadataIndexWriteJob extends AbstractRabbitMqJobHandler {
     // Update the token's metadata
     const result = await idb.oneOrNone(
       `
-        WITH updated_check AS (
```
nofir, line 88: should this be removed?

nofir (changes requested):

PR #6846, opened 2023-11-01
feat: adding new bulk transfer api with txHash array support

JadenDurnford:


georgeroman:
@JadenDurnford Is this actually a backwards-incompatible change? The API stil supports filtering by a single tx hash.

pedromcunha:
Maybe @georgeroman is right and we can continue to use v2.

georgeroman:
@JadenDurnford Looks good, let's just get this to dev first

PR #6878, opened 2023-11-03
feat: remove metadata

fdmota:
Removing env variables from the indexer
@d3or the liquidityOnly flag was function of `METADATA_API_BASE_URL`  which no longer exists. I am not sure what it should be replaced by.

### DIFF
--- a/packages/indexer/src/config/index.ts
+++ b/packages/indexer/src/config/index.ts
@@ -19,14 +19,13 @@ export const config = {
   openseaIndexerApiBaseUrl: String(process.env.OPENSEA_INDEXER_API_BASE_URL),
 
   // When running in liquidity-only mode, all metadata processes are disabled
-  liquidityOnly: !process.env.METADATA_API_BASE_URL,
+  liquidityOnly: true,
   metadataIndexingMethod: String(process.env.METADATA_INDEXING_METHOD || "opensea"),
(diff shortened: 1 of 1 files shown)


packages/indexer/src/config/index.ts
```diff
@@ -19,14 +19,13 @@ export const config = {
   openseaIndexerApiBaseUrl: String(process.env.OPENSEA_INDEXER_API_BASE_URL),
 
   // When running in liquidity-only mode, all metadata processes are disabled
-  liquidityOnly: !process.env.METADATA_API_BASE_URL,
+  liquidityOnly: true,
```
fdmota, line 22: @d3or the liquidityOnly flag was function of `METADATA_API_BASE_URL`  which no longer exists. I am not sure what it should be replaced by.

nofir: this should actually be false by default or you can have:
liquidityOnly: process.env.LIQUIDITY_ONLY (which would be false as its not set anywhere)

nofir: (We actully dont have a usecase currently where "liquidityOnly" is true)

fdmota: ok

nofir (approved):

PR #6882, opened 2023-11-03
feat: add fee recipients

tv3636:


packages/indexer/src/models/fee-recipients/feeRecipients.json
```diff
@@ -98,5 +98,60 @@
     "address": "0xca9337244b5f04cb946391bc8b8a980e988f9a6a",
     "domain": "magiceden.io",
     "kind": "marketplace"
+  },
```
nofir, line 101: this looks ok, did you confirm if the logic will automatically load this into the db? you might need to clean the cache (if its on redis) or add manually

tv3636: yeah confirmed with Idan that this will be loaded into the db

ipeleg (approved):

PR #6888, opened 2023-11-06
feat: add mint refresh delayed jobs

lljxx1:


packages/indexer/src/orderbook/mints/index.ts
```diff
@@ -380,3 +381,18 @@ export const getAmountMintableByWallet = async (
 
   return amountMintable;
 };
+
+export const generateMintRefreshJobIfNeed = async (collection: string) => {
```
georgeroman, line 385: I think this method should belong to the `mints-refresh.ts` job file instead of here.

PR #6968, opened 2023-11-15
feat: replace syncEvents with syncEventsv2

d3or:


packages/indexer/src/api/endpoints/admin/post-sync-events.ts
```diff
@@ -20,22 +19,8 @@ export const postSyncEventsOptions: RouteOptions = {
       "x-admin-api-key": Joi.string().required(),
     }).options({ allowUnknown: true }),
     payload: Joi.object({
```
georgeroman, line 21: Is there a way to keep this logic? Especially the by address syncing - it's much faster to sync 1 million events with  an address filter compared to no address filter.

georgeroman (dismissed):

ipeleg (dismissed):

ipeleg (approved):

PR #6974, opened 2023-11-15
feat: Trending mints patches

r3lays:


packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -315,16 +328,24 @@ async function formatCollections(
 
         tokenCount: Number(metadata.token_count || 0),
         ownerCount: Number(metadata.owner_count || 0),
+        sampleImages:
```
lunarmayor, line 331: i feel like what you really want here is recent sales and not sample images

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -396,6 +401,48 @@ export const getTrendingMints = async (params: {
             field: "pricing.priceDecimal",
           },
         },
+        mints_last_6_hours: {
```
lunarmayor, line 404: kinda wonder if it would be faster to do this as a second test. wouldn't this mean that you have to aggregate and get top hits for collections you won't return

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -396,6 +401,48 @@ export const getTrendingMints = async (params: {
...
+          aggs: {
+            count: {
+              value_count: {
+                field: "id",
+              },
+            },
+          },
+        },
+        top_token_images: {
+          top_hits: {
```
lunarmayor, line 439: found that this type of aggregation makes the whole thing pretty slow since it has to gather all the tokens. how big of an impact is this

packages/indexer/src/elasticsearch/indexes/activities/index.ts
```diff
@@ -396,6 +401,48 @@ export const getTrendingMints = async (params: {
...
+          },
+          aggs: {
+            count: {
+              value_count: {
+                field: "id",
+              },
+            },
+          },
+        },
+        top_token_images: {
```
lunarmayor, line 438: does seem like if we're going to do this we should sort and return more data about the recent sales. otherwise, we should create a more optimized endpoint for returning recent sales

packages/indexer/src/elasticsearch/indexes/activities/index.ts
r3lays: Didn't seem like it was causing a performance hit at all. Tested on Polygon and Maine.

packages/indexer/src/elasticsearch/indexes/activities/index.ts
r3lays: I originally had it separated in another aggregation however I noticed that when it was in the same aggregation there wasn't a noticeable performance hit so I left it in the same query.

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: What exactly is the action item here then? 
To get the page out - we at least need this. Where should we get the sales from?

packages/indexer/src/elasticsearch/indexes/activities/index.ts
r3lays: > i feel like what you really want here is recent sales and not sample images
Similar to this conversation. This is something we could definitely do however will we be getting them from elastic search or fill events?

packages/indexer/src/elasticsearch/indexes/activities/index.ts
lunarmayor: ah okay sounds good

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
lunarmayor: guess let's just roll with this for now. maybe it can just be the case that if you want recent sales you fetch them separately but this gives you the ability to build the sample image ui without much extra effort

lunarmayor: mean recent mints not recent sales

lunarmayor: you could fetch all the trending mints and then for each go get the recent mint activity for them. More just want to figure out what our answer is for building something like this

lunarmayor: Screen Shot 2023-11-21 at 11 20 58 AM
IMG_4182

pretty common to do

lunarmayor (dismissed):

packages/indexer/src/elasticsearch/indexes/activities/index.ts
nofir: regarding top_token_images:
- take into consideration that we don't refresh the token image on the activities so its not available in the time of indexing this will be empty (on the activities apis we use the token cache to get the most up to date version)
- would it make sense to have it as part of the collection cache?

r3lays (commented):
lgtm

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
```diff
@@ -289,10 +304,12 @@ async function formatCollections(
             : null,
         };
       }
+      const sampleImages = _.filter(
```
nofir, line 307: did you see cases where this was failing?

we have a check in Assets.getLocalAssetsLink:

    if (_.isEmpty(assets) || assets == "") {
      return undefined;
    }


not an issue but trying to understand the usecase

packages/indexer/src/api/endpoints/collections/get-trending-collections/v1.ts
```diff
@@ -232,7 +233,15 @@ export async function getCollectionsMetadata(collectionsResult: any[]) {
...
-      
+
+      ARRAY(
+        SELECT
+          tokens.image
+        FROM tokens
+        WHERE tokens.collection_id = collections.id
+        LIMIT 4
+      ) AS sample_images,
+
```
nofir, line 244: In later version of our collections api we do:

`ARRAY(
  SELECT
    tokens.image
  FROM tokens
  WHERE tokens.collection_id = collections.id
  ORDER BY rarity_rank DESC NULLS LAST
  LIMIT 4
) AS sample_images,`

packages/indexer/src/api/endpoints/collections/get-trending-mints/v1.ts
r3lays: Nope,  just wanted to stay consist with the logic I saw in another endpoint.

nofir (approved):

PR #7008, opened 2023-11-17
fix: filter disabled metadata collections on search

tv3636:


nofir (approved):
Please have angel test it on dev before deploying

PR #7024, opened 2023-11-20
deor/platf 2243 add image version updated at

d3or:
- feat: remove any
- feat: rename jobs & add to job list
- feat: if no token uri, write plain metadata
- feat: use onchain indexing for tokens that have slug
- feat: update add to uri queue logic
- feat: add onchain to get-provider-metadata
- feat: fix calling wrong method, not extending
- feat: add better handling for onchain edge cases
- feat: default to token id as name
- feat: add endpoint to increment cache version
- feat: fall back to simplehash
- feat: add metadata index process job when fallback
- feat: cleanup
- feat: rename to image_version_updated_at
- feat: remove to buffer on collection id
- feat: fix route
- feat: rename image_version_updated_at
- feat: only fallback to simplehash if fallbackTokens is not empty
- feat: remove comments
- feat: fix comments, add fallbackMetadataIndexingMethod
- feat: update apis to use image_version_upated_at


packages/indexer/src/api/endpoints/admin/post-update-image-version.ts
```diff
@@ -0,0 +1,74 @@
...
+    const payload = request.payload as any;
+
+    try {
+      // if token is provided, refresh the token
+      if (payload.token) {
+        const [contract, tokenId] = payload.token.split(":");
+        await idb.oneOrNone(
+          `
+          UPDATE tokens
+            SET image_version_updated_at = NOW()
```
nofir, line 48: you need to set the updated_at=now()

packages/indexer/src/api/endpoints/admin/post-update-image-version.ts
```diff
@@ -0,0 +1,74 @@
...
+        `,
+          [toBuffer(contract), tokenId]
+        );
+        return { message: "Request accepted" };
+      }
+
+      const collectionId = payload.collection;
+
+      // update all tokens in the collection
+      await idb.oneOrNone(
```
nofir, line 59: tbh, i would remove it right now and just leave the logic to update a specific token or get the collection from the db and check the token count and if < 10k, allow it. otherwise we can have weird complaints that only some of the token images were invalidated. (we can add an admin api to allow to do it on any collections, similar to the refresh metadata)

packages/indexer/src/api/endpoints/admin/post-update-image-version.ts
```diff
@@ -0,0 +1,74 @@
...
+        `
+        UPDATE tokens
+          SET image_version_updated_at = NOW()
```
nofir, line 62: you need to set the updated_at=now()

packages/indexer/src/api/endpoints/orders/get-user-top-bids/v1.ts
```diff
@@ -282,7 +282,7 @@ export const getUserTopBidsV1Options: RouteOptions = {
             LIMIT 1
         ) y ON TRUE
         LEFT JOIN LATERAL (
-            SELECT t.token_id, t.name, t.image, t.collection_id, floor_sell_value AS "token_floor_sell_value", last_sell_value AS "token_last_sell_value"
+            SELECT t.token_id, t.image_version_updated_at, t.name, t.image, t.collection_id, floor_sell_value AS "token_floor_sell_value", last_sell_value AS "token_last_sell_value"
```
nofir, line 285: this is really not an issue but i realize "image_version_updated_at" is a bit confusing as there is not image version, the timestamp is the version. probably ok to leave as is but a better name would probably be `image_version` which is the timestamp. 

Again, we can ship it as is and it will be fine.

packages/indexer/src/api/endpoints/transfers/get-sales/v4.ts
```diff
@@ -236,7 +236,9 @@ export const getSalesV4Options: RouteOptions = {
                   tokens_data.name,
                   tokens_data.image,
                   tokens_data.collection_id,
-                  tokens_data.collection_name
+                  tokens_data.collection_name,
+                  tokens_data.image_version_updated_at
+
```
nofir, line 241: redundant space

packages/indexer/src/utils/assets.ts
```diff
@@ -30,24 +30,6 @@ export class Assets {
     }
 
     return assets;
```
nofir, line 32: Is the getLocalAssetsLink still being used at all? does it even make sense to keep it (and migrate everywhere to use  getResizedImageUrl)?

nofir (changes requested):

packages/indexer/src/utils/assets.ts
d3or: getLocalAssetsLink is used when we are dealing with an array of images/sample images

nofir (dismissed):

packages/indexer/src/utils/assets.ts
nofir: i think we can still get rid of it but its not a priority for now

packages/indexer/src/api/endpoints/activities/get-collection-activity/v6.ts
```diff
@@ -305,6 +308,7 @@ export const getCollectionActivityV6Options: RouteOptions = {
                       token_id: tokenResult.token_id,
                       name: tokenResult.name,
                       image: tokenResult.image,
+                      image_version: tokenResult.image_version,
```
nofir, line 311: you are missing the cdc logic:
https://github.com/reservoirprotocol/indexer/blob/96be7513f57844350cd2aa80f506942b012b2438/packages/indexer/src/jobs/cdc/topics/indexer-tokens.ts#L48

nofir (changes requested):

packages/indexer/src/jobs/cdc/topics/indexer-tokens.ts
```diff
@@ -53,6 +53,9 @@ export class IndexerTokensHandler extends KafkaEventHandler {
           token_id: payload.after.token_id,
           name: payload.after.name,
           image: payload.after.image,
+          image_version: payload.after?.image_version
+            ? Math.floor(new Date(payload.after.image_version).getTime() / 1000)
```
nofir, line 57: do you still need to do that? (you are doing it in the signImage on line 108)?

nofir (approved):

nofir (dismissed):

nofir (approved):

PR #7075, opened 2023-11-23
feat: add titles.xyz mint

dievardump:
Adds "Custom Mint Integration: Titles.xyz"


packages/indexer/src/orderbook/mints/calldata/detector/titlesxyz.ts
```diff
@@ -0,0 +1,104 @@
...
+    new Interface([
+      "function price() external view returns (uint256)",
+      "function maxSupply() external view returns (uint256)",
+      "function mintLimitPerWallet() external view returns (uint256)",
+      "function saleEndTime() external view returns (uint256)",
+    ]),
+    baseProvider
+  );
+
+  const [price, maxSupply, mintLimitPerWallet, endTime]: [string, string, string, number] =
```
georgeroman, line 50: There's a couple of things that we need to tweak here:
- all of these parameters use 0 as the default value, meaning that they're not checked (for example if the maximum supply or the mint limit per wallet are 0 the checks for them will be skipped)
- the price should include the derivative fee

image

What I usually do as a way to double check that the parsed mint is actually accurate and mintable is run the simulate function as below (the return value will be the success status of the simulation):

image

PR #7101, opened 2023-11-27
feat: update collection image size

d3or:


packages/indexer/src/api/endpoints/activities/get-collection-activity/v6.ts
```diff
@@ -280,9 +280,12 @@ export const getCollectionActivityV6Options: RouteOptions = {
             tokens.image,
             tokens.image_version,
             tokens.metadata_disabled,
+            collections.image_version AS collection_image_version,
             tokens.rarity_score,
             tokens.rarity_rank
           FROM tokens
+          LEFT JOIN collections
```
nofir, line 287: We actually need to update the activities index to include collection_image_version. We need to make sure the following:
- the activities jobs need to support updating the collection_image_version
- we need to trigger a refresh of the activity collection when the collection_image_version changes.

nofir (changes requested):

packages/indexer/src/utils/assets.ts
```diff
@@ -11,7 +11,11 @@ export enum ImageSize {
 }
 
 export class Assets {
-  public static getLocalAssetsLink(assets: string | string[]) {
+  public static getLocalAssetsLink(
```
nofir, line 14: Why do we need "getLocalAssetsLink"? Can't we use the same getResizedImageUrl method? or change it to call the getResizedImageUrl?

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-job.ts
```diff
@@ -431,7 +433,9 @@ export class TokenWebsocketEventsTriggerJob extends AbstractRabbitMqJobHandler {
               {
                 id: r.collection_id,
                 name: r.collection_name,
-                image: r?.collection_image ? Assets.getLocalAssetsLink(r.collection_image) : null,
+                image: r?.collection_image
+                  ? Assets.getResizedImageUrl(r.collection_image, ImageSize.small)
```
nofir, line 437: why small? shouldn't it be medium?

packages/indexer/src/jobs/websocket-events/token-websocket-events-trigger-job.ts
```diff
@@ -161,7 +161,9 @@ export class TokenWebsocketEventsTriggerJob extends AbstractRabbitMqJobHandler {
               {
                 id: data.after.collection_id,
                 name: r?.collection_name,
-                image: r?.collection_image ? Assets.getLocalAssetsLink(r.collection_image) : null,
+                image: r?.collection_image
+                  ? Assets.getResizedImageUrl(r.collection_image, ImageSize.small)
```
nofir, line 165: why small? shouldn't it be medium?

nofir (changes requested):

packages/indexer/src/utils/assets.ts
d3or: its called in around ~20 different places when we are creating an array of images (sample images) usually.

nofir (approved):

nofir (approved):

PR #7112, opened 2023-11-28
feat: add Reddit community tokens

WarmBeer:
I would like to add these Reddit community tokens (All ERC-20 tokens on Polygons) so that I can support them on the [RCAX Reddit Avatar Marketplace](https://marketplace.rcax.io).

### DIFF
--- a/packages/indexer/src/config/network.ts
+++ b/packages/indexer/src/config/network.ts
@@ -683,6 +683,57 @@ export const getNetworkSettings = (): NetworkSettings => {
               name: "RCAX",
               symbol: "RCAX",
               decimals: 18,
+              metadata: {
+                image: "https://rcax.io/images/coins/rcax/icon.png",
+              },
+            },
+          ],
+          [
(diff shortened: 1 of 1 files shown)


WarmBeer:
Reminder 😁

packages/indexer/src/config/network.ts
```diff
@@ -683,6 +683,57 @@ export const getNetworkSettings = (): NetworkSettings => {
               name: "RCAX",
               symbol: "RCAX",
               decimals: 18,
+              metadata: {
+                image: "https://rcax.io/images/coins/rcax/icon.png",
+              },
+            },
+          ],
+          [
+            "0xc6268A296c810024aa3AA2F5Cc2c255bF995AA44",
```
nofir, line 692: all address values need to be in lowercase

nofir (changes requested):

packages/indexer/src/config/network.ts
WarmBeer: Updated now. Thank you for reviewing!

nofir (approved):

PR #7120, opened 2023-11-28
feat: add top traders api

lunarmayor:


packages/indexer/src/api/endpoints/collections/get-top-traders/v1.ts
```diff
@@ -0,0 +1,79 @@
...
+      collection: Joi.string()
+        .lowercase()
+        .required()
+        .description(
+          "Filter to a particular collection, e.g. `0x8d04a8c79ceb0889bdd12acdf3fa9d207ed3ff63`"
+        ),
+    }),
+    query: Joi.object({
+      period: Joi.string()
+        .valid("6h", "1d", "24h", "7d")
```
d3or, line 35: Why 1d and 24h?

d3or (changes requested):

packages/indexer/src/api/endpoints/collections/get-top-traders/v1.ts
lunarmayor: originally ME asked for 24h as a proxy in the trending collections endpoint but i can remove that here.

ipeleg (approved):

PR #7129, opened 2023-11-29
feat: save token uri to tokens table

d3or:


packages/indexer/src/metadata/providers/onchain-metadata-provider.ts
```diff
@@ -145,6 +146,30 @@ export class OnchainMetadataProvider extends AbstractBaseMetadataProvider {
       })
     );
 
+    // Save the tokenURIs to the database
+    await Promise.all(
```
nofir, line 150: Should move to metadata index write job to be written as part of updating the token

packages/indexer/src/models/tokens/index.ts
```diff
@@ -468,4 +468,49 @@ export class Tokens {
       maker: fromBuffer(r.maker),
     }));
   }
+
```
nofir, line 471: i think we can skip these changes for now

nofir (changes requested):

nofir (approved):

PR #7135, opened 2023-11-29
deor/platf 2448 fetch collection metadata from contracturi for all chains, add deployer saving

d3or:
- feat: set contract metadata at deploy (if available)
- fix: dont json stringify

### DIFF
--- a/packages/indexer/src/jobs/collections/collection-contract-deployed.ts
+++ b/packages/indexer/src/jobs/collections/collection-contract-deployed.ts
@@ -102,7 +102,7 @@ export class CollectionNewContractDeployedJob extends AbstractRabbitMqJobHandler
           symbol: symbol || null,
           name: name || null,
           deployed_at: payload.blockTimestamp ? new Date(payload.blockTimestamp * 1000) : null,
-          metadata: collectionMetadata ? JSON.stringify(collectionMetadata) : null,
+          metadata: collectionMetadata ? collectionMetadata : null,
           deployer: deployer ? toBuffer(deployer) : null,
         }
       ),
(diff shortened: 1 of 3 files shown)


packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -76,24 +79,31 @@ export class CollectionNewContractDeployedJob extends AbstractRabbitMqJobHandler
...
           $/name/,
-          $/deployed_at/
+          $/deployed_at/,
+          $/metadata:json/,
+          $/deployer/
         )
         ON CONFLICT (address) DO UPDATE SET
           symbol = EXCLUDED.symbol,
-          name = EXCLUDED.name
+          name = EXCLUDED.name,
```
nofir, line 96: i think we should change this to DO NOTHING

packages/indexer/src/jobs/collections/collection-contract-deployed.ts
```diff
@@ -68,6 +69,8 @@ export class CollectionNewContractDeployedJob extends AbstractRabbitMqJobHandler
 
     const { symbol, name } = await getContractNameAndSymbol(contract);
 
+    const collectionMetadata = await onchainMetadataProvider._getCollectionMetadata(contract);
```
nofir, line 72: change to contractMetadata

packages/indexer/src/models/collections/index.ts
```diff
@@ -231,6 +231,9 @@ export class Collections {
                   FROM collections
                   WHERE collections.id = $/id/
                 ) AS old_metadata
+      UPDATE contracts SET
```
nofir, line 234: not needed

nofir (changes requested):

nofir (approved):

PR #7136, opened 2023-11-29
feat: add metadata jsonb column to contracts table

d3or:


packages/indexer/src/migrations/1700188900743_contract_metadata_column.sql
```diff
@@ -0,0 +1,8 @@
+-- Up Migration
+
+ALTER TABLE contracts ADD COLUMN metadata JSONB
```
nofir, line 3: add deployer

nofir (changes requested):

nofir (approved):

PR #7170, opened 2023-12-01
feat: collections_mints_index

r3lays:


packages/indexer/src/migrations/1701455444389_collection_mints_index.sql
r3lays, line 1: Does it make more sense here to not include the WHERE? In the case where we might need to use this index for other reasons?

ipeleg (approved):

ipeleg (approved):

PR #7267, opened 2023-12-11
feat: limit field sizes

d3or:


packages/indexer/src/config/index.ts
```diff
@@ -21,6 +21,7 @@ export const config = {
   // When running in liquidity-only mode, all metadata processes are disabled
   liquidityOnly: Boolean(Number(process.env.LIQUIDITY_ONLY)),
   metadataIndexingMethod: String(process.env.METADATA_INDEXING_METHOD || "opensea"),
+  metadataMaxFieldSize: Number(process.env.METADATA_MAX_FIELD_SIZE || 1),
```
nofir, line 24: is this in mb? maybe we can be more explicit metadataMaxFieldSizeMB / metadataMaxFieldSizeKB

packages/indexer/src/metadata/providers/utils.ts
```diff
@@ -182,3 +184,27 @@ export class CollectionNotFoundError extends Error {
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function limitFieldSize(value: any) {
+  if (typeof value === "string") {
+    logger.info(
+      "limitFieldSize",
+      JSON.stringify({
+        size: new TextEncoder().encode(value).length,
+      })
+    );
+    return new TextEncoder().encode(value).length > config.metadataMaxFieldSize * 1024 * 1024
```
nofir, line 197: are there any considerations as far as efficiency with encoding the value? i see this: https://stackoverflow.com/questions/9864662/how-to-get-the-string-length-in-bytes-in-nodejs
could be that its the same effort

packages/indexer/src/metadata/providers/utils.ts
```diff
@@ -182,3 +184,27 @@ export class CollectionNotFoundError extends Error {
...
+      ? null
```
nofir, line 198: im wondering if we should just log in the start before but if the env var is set high so it should be ok

packages/indexer/src/config/index.ts
nofir: or at least leave a comment

packages/indexer/src/metadata/providers/utils.ts
d3or: i think it should be fine, both methods convert to utf8 to find the bytes, i dont expect one or the other to have a significant difference

nofir (dismissed):

nofir (approved):

nofir (approved):

PR #7341, opened 2023-12-14
feat: add migration (classifying airdrops

d3or:


packages/indexer/src/migrations/17025792250000_fill_events_kind.sql
```diff
@@ -0,0 +1,7 @@
+-- Up Migration
+
+ALTER TYPE "fill_events_2" ADD COLUMN "kind" TEXT;
```
georgeroman, line 3: I think `TEXT` is wasteful vs using an enum as we do in other places like `order_type_t`.

nofir, line 3: we usually create an enum:
```
CREATE TYPE "order_kind_t" AS ENUM (
  'wyvern-v2',
  'wyvern-v2.3',
  'looks-rare',
  'opendao-erc721',
  'opendao-erc1155',
  'zeroex-v4-erc721',
  'zeroex-v4-erc1155',
  'foundation',
  'x2y2',
  'seaport'
)
```;

**Review by nofir — CHANGES_REQUESTED** [review]

File: packages/indexer/src/migrations/17025792250000_fill_events_kind.sql [hunk 2]
```diff
@@ -0,0 +1,14 @@
+-- Up Migration
+CREATE TYPE "fill_event_kind" AS ENUM (
```
georgeroman, line 2: Do we really need transfer and sale here? Maybe for efficiency we can represent transfers as null (no need to backfill) while for sales from previous experience that's very hard to associate.

nofir (dismissed):

nofir (dismissed):

georgeroman (approved):

PR #7372, opened 2023-12-15
Add collection source to Trending Mints

r3lays:


georgeroman:
@r3lays Looks good! I remember you mentioned some performance issues with the left join, is it fast enough in the end?

georgeroman (approved):

r3lays:
> @r3lays Looks good! I remember you mentioned some performance issues with the left join, is it fast enough in the end?

Yeah it seems to be fine. We have to revamp this EP anyways soon because of an issue with Zora.

PR #7393, opened 2023-12-18
Update network.ts

dturska12:
# Pull Request Template

Request support for $AURA token

- **Code Modifications:**

## Additional Notes

### DIFF
--- a/packages/indexer/src/config/network.ts
+++ b/packages/indexer/src/config/network.ts
@@ -460,6 +460,19 @@ export const getNetworkSettings = (): NetworkSettings => {
               },
             },
           ],
+          [
+            "0x2bff8ddbc1F13F6F976a8f4d7feE677272fB6e0e",
+            {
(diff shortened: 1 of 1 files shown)


packages/indexer/src/config/network.ts
```diff
@@ -460,6 +460,19 @@ export const getNetworkSettings = (): NetworkSettings => {
               },
             },
           ],
+          [
+            "0x2bff8ddbc1F13F6F976a8f4d7feE677272fB6e0e",
```
nofir, line 464: need to be lowercase (also the contract below)

nofir (changes requested):

packages/indexer/src/config/network.ts
dturska12: ah sorry completely forgot about that. Have updated the address

nofir (approved):

PR #7404, opened 2023-12-18
feat: adding oracleEnabled to marketplace configurations

JadenDurnford:
- Added oracleEnabled to marketplace config exchanges

### DIFF
--- a/packages/indexer/src/api/endpoints/collections/get-collection-marketplace-configurations/v1.ts
+++ b/packages/indexer/src/api/endpoints/collections/get-collection-marketplace-configurations/v1.ts
@@ -53,6 +53,7 @@ type Marketplace = {
       supportedBidCurrencies: string[];
       maxPriceRaw?: string;
       minPriceRaw?: string;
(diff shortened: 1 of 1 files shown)


georgeroman:
@JadenDurnford We recently added support for oracle orders via payment-processor-v2 so we should also set that field to true for the payment-processor-v2 exchange (and it works on all chains, unlike the seaport one).

JadenDurnford:
> @JadenDurnford We recently added support for oracle orders via payment-processor-v2 so we should also set that field to true for the payment-processor-v2 exchange (and it works on all chains, unlike the seaport one).

@georgeroman I believe I added that here: https://github.com/reservoirprotocol/indexer/pull/7404/files#diff-e87fcd3853f72ee7fa078a632e699120548242807de170d7eb9978b90927f41fR303, is there anywhere else I need to set the field to true?

georgeroman:
Great, that should be enough

pedromcunha (approved):

PR #7468, opened 2023-12-20
feat: upload data: uri images to cloudflare

d3or:


packages/indexer/src/jobs/metadata-index/metadata-image-upload-job.ts
```diff
@@ -0,0 +1,104 @@
...
+    const { imageURI, contract, tokenId } = payload;
+
+    try {
+      let imageBuffer;
+
+      if (this.isDataURI(imageURI)) {
+        // Handle data URI
+        imageBuffer = this.dataURIToBuffer(imageURI);
+      } else {
+        // Handle standard image URL
```
nofir, line 34: Why is that needed?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
```diff
@@ -34,33 +35,15 @@ export default class OnchainMetadataProcessTokenUriJob extends AbstractRabbitMqJ
...
-            metadata[0].imageUrl = null;
-          }
+          await metadataImageUploadJob.addToQueue({
+            contract,
+            tokenId,
+            imageURI: metadata[0].imageUrl,
+          });
+          metadata[0].imageUrl = null;
+        } else {
+          metadata[0].imageUrl = null;
```
nofir: why are you nullfying it?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
```diff
@@ -34,33 +35,15 @@ export default class OnchainMetadataProcessTokenUriJob extends AbstractRabbitMqJ
...
-            return;
-          } else {
-            metadata[0].imageUrl = null;
-          }
+          await metadataImageUploadJob.addToQueue({
+            contract,
+            tokenId,
+            imageURI: metadata[0].imageUrl,
+          });
+          metadata[0].imageUrl = null;
```
nofir, line 43: I realize that we might a race condition with the metadataIndexWriteJob where the image upload will finish and set the image url and then the metadataIndexWriteJob updates the token

nofir (changes requested):

packages/indexer/src/jobs/metadata-index/metadata-image-upload-job.ts
```diff
@@ -0,0 +1,130 @@
...
+import { logger } from "@/common/logger";
+import { config } from "@/config/index";
+import { Tokens } from "@/models/tokens";
+import { Collections } from "@/models/collections";
+
+export type MetadataImageUploadJobPayload = {
+  tokenId: string;
+  contract: string;
+  imageURI: string;
+  kind: "token-image" | "token-media" | "token-uri" | "collection-image" | "collection-banner";
```
nofir, line 13: I would do something like this:
https://github.com/reservoirprotocol/indexer/blob/3af8d4aa78e598e47b9d4227d0826c4ef6c33ee0/packages/indexer/src/jobs/elasticsearch/activities/process-activity-event-job.ts#L19

packages/indexer/src/jobs/metadata-index/metadata-image-upload-job.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (!uploadResult.success) {
+        throw new Error(`Failed to upload to Cloudflare: ${uploadResult.errors}`);
+      }
+
+      const originalUrl = uploadResult.result.variants[2];
+
+      switch (payload.kind) {
+        case "token-image":
+          await Tokens.update(contract, tokenId, {
+            image: originalUrl,
```
nofir, line 51: do you know if we upload two identical images, does it still create 2 different images on cloudflare?

packages/indexer/src/jobs/metadata-index/metadata-write-job.ts
```diff
@@ -105,15 +105,15 @@ export default class MetadataIndexWriteJob extends AbstractRabbitMqJobHandler {
           name = $/name/,
           token_uri = $/tokenURI/,
           description = $/description/,
-          image = $/image/,
+          image = CASE WHEN $/image/ IS NOT NULL THEN $/image/ ELSE image END,
```
nofir, line 108: https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-coalesce/

packages/indexer/src/jobs/metadata-index/metadata-image-upload-job.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    if (!response.ok) {
+      throw new Error(`Cloudflare API responded with ${response.status}: ${response.statusText}`);
+    }
+
+    return response.json();
+  }
+
+  private async checkIfImageExistsInCloudflare(hash: string) {
+    const url = `https://api.cloudflare.com/client/v4/accounts/${config.cloudflareAccountID}/images/v1/${hash}`;
```
nofir, line 195: do we have a different account between dev / prod or all images are mixed?

packages/indexer/src/jobs/metadata-index/onchain-metadata-process-token-uri-job.ts
```diff
@@ -33,78 +35,41 @@ export default class OnchainMetadataProcessTokenUriJob extends AbstractRabbitMqJ
...
-          }
+              },
+            ],
+            true,
+            5
+          );
+          return;
         }
 
+        await this.uploadImageIfNecessary(metadata[0]);
```
nofir, line 64: does it make sense to have this in the abstract metadata provider?
also, i thought we discussed yesterday to deploy this but not enable it using an env var, not sure i see where we do that

nofir (changes requested):

PR #7595, opened 2024-01-04
feat: use redis to debug load testing results

tv3636:


packages/indexer/src/api/endpoints/orders/get-orders-asks/v4.ts
```diff
@@ -609,10 +625,16 @@ export const getOrdersAsksV4Options: RouteOptions = {
         );
       }
 
-      return {
+      const response = {
         orders: await Promise.all(result),
         continuation,
       };
+
+      if (process.env.LOAD_TEST_KEY && apiKey && apiKey.key === process.env.LOAD_TEST_KEY) {
```
nofir, line 633: This means that we save it to redis even if its already on redis , no?

Shouldn't this happen in after line 539? so you would get / set to redis only the response from the db and then the rest is the same. 

Not critical as its only for doing the load test so im fine if you want to deploy it as is

tv3636: If it already got the result from redis it wouldn't reach here

and yeah I was first trying to set it at line 539 but it got tricky handling buffer values converting from JSON to string and back. For the purposes of testing it shouldn't matter since we already know the time after the db query returns is insignificant

nofir: fair enough

nofir (dismissed):

nofir (approved):

PR #7604, opened 2024-01-04
feat: more fixes

d3or:


packages/indexer/src/api/endpoints/tokens/get-tokens/v6.ts
```diff
@@ -1414,7 +1415,14 @@ export const getTokensV6Options: RouteOptions = {
           }
         }
 
-        const metadata = parseMetadata(r);
+        const metadata = parseMetadata(r.metadata);
+        if (!r.image && r.metadata?.image_original_url) {
```
nofir, line 1419: why not put this inside the parseMetadata method?

nofir (approved):

PR #7630, opened 2024-01-08
Deor/platf 2953 return airdrop data on transfers websocket

d3or:


packages/indexer/src/jobs/websocket-events/transfer-websocket-events-trigger-job.ts
```diff
@@ -40,6 +40,7 @@ export class TransferWebsocketEventsTriggerQueueJob extends AbstractRabbitMqJobH
         txHash: data.after.tx_hash,
         logIndex: data.after.log_index,
         batchIndex: data.after.batch_index,
+        isAirdrop: data.after.kind === "airdrop",
```
d3or, line 43: Add enum

PR #7647, opened 2024-01-09
Spam Collection evaluation queue updates

d3or:


packages/indexer/src/jobs/collections-refresh/collections-check-spam-job.ts
```diff
@@ -12,6 +12,7 @@ import { logger } from "@/common/logger";
 
 export type CollectionCheckSpamJobPayload = {
   collectionId: string;
+  trigger: "default" | "metadata-changed" | "transfer-burst";
```
nofir, line 15: default can be removed

packages/indexer/src/jobs/collections-refresh/collections-check-spam-job.ts
```diff
@@ -21,10 +22,20 @@ export default class CollectionCheckSpamJob extends AbstractRabbitMqJobHandler {
...
     const collection = await Collections.getById(collectionId, true);
 
     if (collection) {
+      // if the spam was manually set by a trusted partner, don't change it
+      if (collection.isSpam === 100 || collection.isSpam === -100) {
+        return;
+      }
+
+      // if the collection is verified and marked as not spam -> do nothing
+      if (collection.metadata?.safelistRequestStatus === "verified" && collection.isSpam < 0) {
```
nofir, line 35: i think it should be collection.isSpam =< 0,  if verified by opensea, we can skip the rest (even if its not an override)

nofir (changes requested):

nofir (approved):

PR #7659, opened 2024-01-10
feat: extend punks2023 metadata

haseebahmad109:


packages/indexer/src/metadata/extend/punks2023/index.ts
```diff
@@ -0,0 +1,21 @@
+import axios from "axios";
+import { TokenMetadata } from "@/metadata/types";
+
+const metadataBaseURI = "https://punks2023.com/api/marketplace";
```
ipeleg, line 4: @haseebahmad109 let's move this const to be inside the extend function

packages/indexer/src/metadata/extend/punks2023/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const extend = async (metadata: TokenMetadata) => {
+  const response = await axios.get(`${metadataBaseURI}/${metadata.tokenId}`);
+
+  const { attributes: customAttributes } = response.data;
```
ipeleg, line 9: What is customAttributes? shouldn't it be defined?

ipeleg (changes requested):

packages/indexer/src/metadata/extend/punks2023/index.ts
haseebahmad109: @ipeleg 
Fixed and moved to extend function.

packages/indexer/src/metadata/extend/punks2023/index.ts
haseebahmad109: @ipeleg 

Type for customAttributes is now defined.

ipeleg (approved):

PR #7814, opened 2024-01-24
feat: pending tx ws

lljxx1:
# Pull Request Template

- **Code Modifications:**

## Additional Notes


packages/indexer/src/jobs/websocket-events/utils.ts
```diff
@@ -1,5 +1,29 @@
+import { idb } from "@/common/db";
+import { toBuffer } from "@/common/utils";
+
 // Utility functions for parsing cdc event data
 
+export async function getTokenMetadata(tokenId: string, contract: string) {
```
nofir, line 6: Is there a reason to create a new method and not use [this](https://github.com/reservoirprotocol/indexer/blob/9965b8b1b7f0f6f7ffc242f5fd7297afadb8dd0e/packages/indexer/src/models/tokens/index.ts#L27)?

packages/indexer/src/tests/pending-state/pending.test.ts
```diff
@@ -22,15 +21,16 @@ describe("PendingState", () => {
       // console.log("pendingTokens", pendingTokens);
       if (!pendingTokens?.length) continue;
       const contract = pendingTokens[0].contract;
-      const pendingTokenIdsBefore = await getContractPendingTokenIds(contract);
-      const recent = await getRecentPendingTokens();
+      const pendingTokenIdsBefore = await getPendingItems(contract);
+      const recent = await getPendingItems();
+      // console.log("recent", recent);
```
nofir, line 26: please remove comment

packages/indexer/src/jobs/websocket-events/utils.ts
nofir: Might be simpler with the mocking but if not, then i would just use the existing one

nofir (changes requested):

nofir (dismissed):

ipeleg (dismissed):

georgeroman:
@nofir or @ipeleg Asking for on last round of review before merging this

ipeleg (approved):

PR #7864, opened 2024-01-29
fix: add multiple fees check

lljxx1:
# Pull Request Template

- **Code Modifications:**

## Additional Notes

### DIFF
--- a/packages/indexer/src/api/endpoints/collections/get-collection-marketplace-configurations/v1.ts
+++ b/packages/indexer/src/api/endpoints/collections/get-collection-marketplace-configurations/v1.ts
@@ -44,6 +44,7 @@ type Marketplace = {
       traitBidSupported: boolean;
       orderKind: OrderKind | null;
       customFeesSupported: boolean;
+      multipleFeesSupported: boolean;
       collectionBidSupported?: boolean;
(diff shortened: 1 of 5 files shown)


packages/indexer/src/api/endpoints/collections/get-collection-marketplace-configurations/v1.ts
```diff
@@ -44,6 +44,7 @@ type Marketplace = {
       traitBidSupported: boolean;
       orderKind: OrderKind | null;
       customFeesSupported: boolean;
+      multipleFeesSupported: boolean;
```
georgeroman, line 47: Can we rename this to `numFeesSupported` instead? It would represent the number of fees the protocol supports (set to `undefined` it unlimited fees are supported)

georgeroman: Also, we only want to set it for Reservoir-supported exchanges (anything below `// Handle Reservoir`)

georgeroman (changes requested):
As discussed, initially let's just log and don't throw any errors. We want to see if anyone is actually using this logic.

PR #7945, opened 2024-02-08
Devo 163 fix cache control

fdmota:
## Description

This PR intends to quickly fix the cache-control directive in the indexer.
We are using 3 different ways of setting the cache that should be harmonized but outside of the scope of this PR.
There was 2 issues:
- the syntax was incorrect `cache-control: 1000`. Correct form is something like `cache-control: max-age=10, must-revalidate, public`
- In the other 2 ways of setting the cache, it uses the library hapi that expects the value to be passed in milliseconds but is converted in seconds under the hood. In this case, where we override the directly the `cache-control` header the value must be in seconds as defined by the [http standard](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). This second issue might the main reason why the cache did not expired in Cloudfront. We were expecting 1 second while it was 1000 seconds (if it was not ignored completely because of the first issue).

## Checklist:

- [ x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published in downstream modules. -->

### DIFF
--- a/packages/indexer/src/api/endpoints/tokens/get-user-tokens/v9.ts
+++ b/packages/indexer/src/api/endpoints/tokens/get-user-tokens/v9.ts
@@ -1027,7 +1027,7 @@ export const getUserTokensV9Options: RouteOptions = {
           tokens: await Promise.all(result),
           continuation,
         })
-        .header("cache-control", `${cacheControl}`);
+        .header("cache-control", `max-age=${cacheControl}, must-revalidate, public`);
     } catch (error) {
       logger.error(`get-user-tokens-${version}-handler`, `Handler failure: ${error}`);
       throw error;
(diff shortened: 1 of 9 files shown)


packages/indexer/src/api/endpoints/tokens/get-user-tokens/v9.ts
```diff
@@ -1017,9 +1017,9 @@ export const getUserTokensV9Options: RouteOptions = {
         };
       });
 
-      let cacheControl = 1000;
+      let cacheControl = 1;
       if (query.includeTopBid || query.includeAttributes) {
-        cacheControl = 1000 * 60;
+        cacheControl = 60;
```
nofir, line 1022: don't you need to fix the .header("cache-control", `${cacheControl}`); on line 1030?

nofir (approved):

PR #8018, opened 2024-02-14
Adding GCOIN to polygon

Nike682631:


nofir:
@Nike682631 I see this token is [supported by coingecko](https://www.coingecko.com/en/coins/galaxy-fight-club). In this case, there is no need to add a custom token as it will be supported

PR #8022, opened 2024-02-14
feat: adding SignedZone and SignedZoneController to opbnb

JadenDurnford:


packages/contracts/hardhat.config.ts
```diff
@@ -286,6 +287,14 @@ const config: HardhatUserConfig = {
           browserURL: "https://scan.ancient8.gg",
         },
       },
+      {
+        network: "opBnb",
```
ipeleg, line 291: let's change this to all lower case opbnb

ipeleg (approved):

PR #8175, opened 2024-02-27
Ofir/platf 3227 queues errors anomaly alarms

nofir:
# Pull Request Template

- **Code Modifications:**

## Additional Notes

### DIFF
--- a/packages/indexer/src/jobs/abstract-rabbit-mq-job-handler.ts
+++ b/packages/indexer/src/jobs/abstract-rabbit-mq-job-handler.ts
@@ -140,11 +140,21 @@ export abstract class AbstractRabbitMqJobHandler {
       }
 
       // Log the error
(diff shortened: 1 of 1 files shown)


packages/indexer/src/jobs/abstract-rabbit-mq-job-handler.ts
```diff
@@ -140,11 +140,23 @@ export abstract class AbstractRabbitMqJobHandler {
       }
 
       // Log the error
-      if (!this.disableErrorLogs || queueName === this.getDeadLetterQueue()) {
+      if (!this.disableErrorLogs && queueName !== this.getDeadLetterQueue()) {
```
ipeleg, line 143: let's add in the message also the `queueName` which can be either the regular queue, or the dead letter

PR #8244, opened 2024-03-06
Jaden/platf 3282 return token floorask on user tokens currently just

JadenDurnford:


packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
```diff
@@ -0,0 +1,1239 @@
...
+import { parseMetadata } from "@/api/endpoints/tokens/get-user-tokens/v8";
+import { MaxUint256 } from "@ethersproject/constants";
+
+const version = "v10";
+
+export const getUserTokensV10Options: RouteOptions = {
+  description: "User Tokens",
+  notes:
+    "Get tokens held by a user, along with ownership information such as associated orders and date acquired.",
+  tags: ["api", "Tokens"],
```
nofir, line 43: you need to mark v9 as deprecated

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
```diff
@@ -0,0 +1,1239 @@
...
+                            },
+                          },
+                          topBidCurrency,
+                          query.displayCurrency
+                        )
+                      : null,
+                    source: getJoiSourceObject(topBidSource),
+                  }
+                : undefined,
+              floorAsk: {
```
nofir, line 1121: just to confirm,  it looks like its using the same values as the ownership-> floorAsk. Wouldn't it mean that it would be available only if the user owns it? (for erc1155 it could be a different floor ask)

nofir (changes requested):

nofir (approved):

PR #8262, opened 2024-03-08
Run db migration inside code

fdmota:
# Run db migration in code

## Description

Kubernetes restart are caused by the db migration failing because they are running in parallel.

## Checklist:

- [x ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published in downstream modules. -->

### DIFF
--- a/packages/indexer/src/common/db-migrate.ts
+++ b/packages/indexer/src/common/db-migrate.ts
@@ -16,7 +16,7 @@ export const runDBMigration = async () => {
       logger.info("postgresql-migration", `Start postgresql migration`);
       try {
         await migrationRunner({
-          dryRun: true,
+          dryRun: false,
           databaseUrl: {
             connectionString: config.databaseUrl
           },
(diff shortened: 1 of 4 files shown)


packages/indexer/src/common/db-migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+import { logger } from "@/common/logger";
+import { delay } from "@/common/utils";
+
+import { config } from "@/config/index";
+
+export const runDBMigration = async () => {
+  const EXPIRATION_LOCK = 300
+  const CHECK_MIGRATION_INTERVAL = 1000
+  const dbMigrationLock = "db-migration-lock";
+  const dbMigrationStatus = "db-migration-status";
```
ipeleg, line 12: I think a better name for this var would be `dbMigrationVersion`

ipeleg (approved):

PR #8283, opened 2024-03-12
Jaden/platf 3319 fr getusertokens api addition

JadenDurnford:


packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
```diff
@@ -747,7 +751,7 @@ export const getUserTokensV10Options: RouteOptions = {
                ${selectLastSale}
                top_bid_id, top_bid_price, top_bid_value, top_bid_currency, top_bid_currency_price, top_bid_currency_value, top_bid_source_id_int,
                o.currency AS collection_floor_sell_currency, o.currency_price AS collection_floor_sell_currency_price, o.currency_value AS collection_floor_sell_currency_value, o.token_set_id AS collection_floor_sell_token_set_id,
-               c.name as collection_name, con.kind, con.symbol, c.metadata, c.royalties, (c.metadata ->> 'safelistRequestStatus')::TEXT AS "opensea_verification_status",
+               c.name as collection_name, c.token_count as collection_token_count, con.kind, con.symbol, extract(epoch from con.deployed_at) AS contract_deployed_at, c.metadata, c.royalties, (c.metadata ->> 'safelistRequestStatus')::TEXT AS "opensea_verification_status",
```
nofir, line 754: remind me, any reason to do `extract(epoch`?

nofir: I think you can just return con.deployed_at and then skip the * 1000 in line 1045

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
```diff
@@ -1046,6 +1053,7 @@ export const getUserTokensV10Options: RouteOptions = {
                 isNsfw: Number(r.c_nsfw_status) > 0,
                 metadataDisabled: Boolean(Number(r.c_metadata_disabled)),
                 openseaVerificationStatus: r.opensea_verification_status,
+                tokenCount: String(r.collection_token_count),
```
nofir, line 1056: do we return it as string in the collections / user collections api?

nofir (changes requested):

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
JadenDurnford: I copied the logic from collections v7: https://github.com/reservoirprotocol/indexer/blob/de3a079602f4922603c613f572caa8bd4c174b69/packages/indexer/src/api/endpoints/collections/get-collections/v7.ts#L739 is it fine to leave it as is or should I still change it?

packages/indexer/src/api/endpoints/tokens/get-user-tokens/v10.ts
JadenDurnford: Yup, it is done like that on the collections api https://github.com/reservoirprotocol/indexer/blob/de3a079602f4922603c613f572caa8bd4c174b69/packages/indexer/src/api/endpoints/collections/get-collections/v7.ts#L854

nofir (approved):

PR #8295, opened 2024-03-13
feat: adding apex testnet contracts

JadenDurnford:


packages/indexer/src/config/network.ts
```diff
@@ -1833,6 +1836,38 @@ export const getNetworkSettings = (): NetworkSettings => {
         },
       };
     }
+    // Apex Testnet
+    case 70800: {
+      return {
+        ...defaultNetworkSettings,
```
tv3636, line 1842: need to add `isTestnet: true`

tv3636 (changes requested):

tv3636 (approved):

PR #8313, opened 2024-03-18
chore: remove lodash imports

RobbyUitbeijerse:
Hello there! We are working with a forked version of Reservoir. For some reason @nofir added a few randomness based conditions half a year ago that cause logs only to come through 75% of the time. That seems like a bad practice, because if you're actively using logs to debug certain situations you're not able to find them.

I strongly feel that these conditions shouldn't be there. It makes debugging things horribly ambiguous. If the Reservoir team feels strongly about keeping the randomness factor in, I would be happy to adjust it to make it configurable based on an environment variable. Let me know if that would be preferred here.


nofir:
@RobbyUitbeijerse the randomness factor was added due to the impact on the cost of our logs solution given the amount of logs that this was producing. To be fair, this might not be an issue on your end so I would suggest you introduce an env var so we can control the level of randomness (it should default to the current value so we would not have any impact).

PR #8329, opened 2024-03-19
Idan/platf 3355 support azuki custom metadata reterival

ipeleg:


packages/indexer/src/metadata/custom/azuki/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    return {};
+  }
+
+  let image = `${config.ordinalsMetadataUrl}/content/${info?.inscriptionId}`;
+  if (info?.data?.image && info.data.image !== uri) {
+    const imageInfo = await fetchSatInformation(info.data.image);
+    image = `${config.ordinalsMetadataUrl}/content/${imageInfo?.inscriptionId}`;
+  }
+
+  return {
```
nofir, line 57: Why is that needed? i think the logic that calls fetchTokenUriMetadata then calls the parseToken logic.

packages/indexer/src/metadata/custom/azuki/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      )}`
+    );
+    return {};
+  }
+
+  if (_.isEmpty(info)) {
+    return {};
+  }
+
+  let image = `${config.ordinalsMetadataUrl}/content/${info?.inscriptionId}`;
```
nofir, line 51: wouldnt it mean that you set an image even if `info?.data?.image` is null?

packages/indexer/src/metadata/custom/azuki/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      "azuki-custom",
+      `failed to get info from uri ${uri} for ${contract}:${tokenId} error = ${JSON.stringify(
+        error
+      )}`
+    );
+    return {};
+  }
+
+  if (_.isEmpty(info)) {
+    return {};
```
nofir, line 48: what would happen if we return empty obj? do we ever do that anywhere?

PR #8457, opened 2024-04-08
Add the address for native USDC on Base Sepolia

jpetrich:
# Description

Add the address for native USDC on Base Sepolia, similarly to how it is defined for other mainnets and testnets.


packages/sdk/src/common/addresses.ts
```diff
@@ -113,6 +113,9 @@ export const Usdc: ChainIdToAddressList = {
     "0xd9aaec86b65d86f6a7b5b1b0c42ffa531710b6ca", // Bridged
     "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913", // Native
   ],
+  [Network.BaseSepolia]: [
+    "0x036CbD53842c5426634e7929541eC2318f3dCF7e", // Native
```
georgeroman, line 117: We always use lowercased addresses

PR #8543, opened 2024-04-24
chore/add zed weth

rplusq:
- chore: add zed custom tokens


packages/indexer/src/config/network.ts
```diff
@@ -1250,19 +1250,35 @@ export const getNetworkSettings = (): NetworkSettings => {
...
             {
-              contract: "0x8ef069e433022cf3625a086096c3d9ec96da0160",
+              contract: "0xAb509Da7810B38317fe9710C57fFCea9c5877a36",
               name: "ZED Token",
               symbol: "ZED",
               decimals: 18,
               metadata: {
                 image:
                   "https://bafkreidcljrhz7hq4h5rarxgxd63tail24mqawllwhxfrvr3esfv2nvyiy.ipfs.nftstorage.link/",
+                  coingeckoCurrencyId: "zed-run"
```
nofir, line 1262: is this from coingecko? i wasn't able to confirm they support it. otherwise, we can just remove coingeckoCurrencyId

packages/indexer/src/config/network.ts
```diff
@@ -1250,19 +1250,35 @@ export const getNetworkSettings = (): NetworkSettings => {
...
+            "0xeB5667fb4b6dFebEB16874B9F26deA002fcEFC3f",
+            {
+              contract: "0xeB5667fb4b6dFebEB16874B9F26deA002fcEFC3f",
+              name: "Wrapped Ether",
+              symbol: "WETH",
+              decimals: 18,
+              metadata: {
+                image:
+                  "https://assets.coingecko.com/coins/images/2518/large/weth.png?1628852295",
+                coingeckoCurrencyId: "ethereum"
```
nofir, line 1277: is this from coingecko? i wasn't able to confirm they support it. otherwise, we can just remove coingeckoCurrencyId

packages/indexer/src/config/network.ts
rplusq: They don't support our testnet contracts, but they have the prod ones, I guess this could help with getting the right price?

packages/indexer/src/config/network.ts
rplusq: They don't support our testnet contracts, but they have the prod ones, I guess this could help with getting the right price?

packages/indexer/src/config/network.ts
nofir: this is intended for custom currency that don't have a value, we can test it out but im not sure if we actually get the price.

rplusq: Let's try it then

nofir: one thing i noticed, contract address needs to be lowercase (in both entries)

nofir (changes requested):

packages/indexer/src/config/network.ts
rplusq: Done

nofir (approved):

